PR #51, opened 2015-01-16
Vhdl range func

orsonmmz:


steveicarus:
This pull request includes a commit that changes the vvp_darray_atom template that I don't like, and I've attached a comment to that commit. I think we want to rework what that commit does. Can you remove just that commit from the pull? Alternatively, we can hold off until CaryR implements a vvp_darray_bool class to solve this issue for you.

orsonmmz:
Ok, no problem. I think we may wait, unless there are changes you would like to merge now. The reason why I proposed the changes is to avoid too wide branch diversion/work replication, as Cary works on dynamic arrays too.
Just out of curiosity - do you see there functionality flaws, or the solution itself or its performance is the show stopper?

PR #70, opened 2015-05-29
Time expressions for vhdlpp

orsonmmz:


steveicarus:
Given our conversation on the mailing list, is it OK if we not include the "... time type" patch in this pull request? I think the conclusion is that time values do not transfer well across scopes, and also making time not a reg type may lead to some subtle issues elsewhere.

orsonmmz:
I know it seems wrong to transfer time through ports, but the code I need to simulate does it and it cannot be changed. I need to think of another way for "input time" ports or find paragraphs that will make the patch legitimate. Would you withhold the pull request for a while?

steveicarus:
Can do.

BTW since "time"-ness doesn't really mean anything, you can use 64bit wires to carry time values if you want, at least for the purposes of your tests.

Anyhow, drop me a note when you are ready.

PR #82, opened 2015-10-22
Make a few constructors explicit.

chrta:
This prevents implicit conversions and a removes few cppcheck warnings.

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -4030,7 +4030,7 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
 	// If the input is a string, and the part select is working on
 	// byte boundaries, then make the result into a string.
       if (par_val.is_string() && (labs(lsv)%8 == 0) && (wid%8 == 0))
-	    return result.as_string();
+	    return verinum(result.as_string());
 
       return result;
 }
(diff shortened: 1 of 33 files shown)


elab_expr.cc
```diff
@@ -4030,7 +4030,7 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
 	// If the input is a string, and the part select is working on
 	// byte boundaries, then make the result into a string.
       if (par_val.is_string() && (labs(lsv)%8 == 0) && (wid%8 == 0))
-	    return result.as_string();
+	    return verinum(result.as_string());
```
chrta: I am not sure about this change. This code is executed in the current master branch, too. But since result is already of type verinum, a "return result;" should work, too?

caryr: From a type perspective that would be correct, but from a functional standpoint it would not be. We created a numeric verinum at the beginning of the routine and that is being converted to a string verinum at this line. The string verinum constructor is the only one that can set the string type to true. Using just "return result" would return a numeric value which is incorrect in the context of a parameter wanting a string value.

Everything else looks good and takes something off my to do list.

Thanks

chrta: Thanks for the explanation.
So no changes are needed to this pull request, right?

caryr: If the pull request has return verinum(result.as_string()); then yes it should be fine.

PR #114, opened 2016-07-07
Changes necessary to compile with gcc on Debian/Linux testing

wzab:
To be able to compile the newest sources with gcc on my Debian/Linux testing machine, I had to modify two files.
Otherwise I get the errors:
 error: ‘ExpRange::range_dir_t’ is not a class or namespace

The other solution could be to compile with "-std=c++0x" option, which should add scoped enums.
I have not tested it though.


martinwhitaker:
Thanks.

PR #116, opened 2016-07-18
Conda fix

tfors:
This PR uses /proc/self/exe (if it exists) on UNIX systems to calculate ivl_root at run-time and falls back to using IVL_ROOT from the Makefile if not.  This allows iverilog to work with the conda package manager which uses patchelf on linux to make its binaries relocatable.


martinwhitaker:
This looks robust enough, so I've merged it, along with a couple of minor fixes.

PR #135, opened 2016-12-07
Fixed cross-compile using mingw 64 bit under Windows 10 Linux subsyst…

udif:
…em (Ubunto 14.04)

Should work with true Ubuntu 14.04 as-is.

Test with:
apt-get install g++-mingw-w64
./configure --host=x86_64-w64-mingw32


steveicarus:
This is not quite right. The "dosify.exe" program, and those like it, should be build with the $(HOSTCC) compiler so that it can be executed on the build system. That is why there is a @HOSTCC@ substitution. Is there a reason you changed that?

udif:
That's right, the problem is the definition of 'host':
https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html

> --build=build-type
the type of system on which the package is being configured and compiled. It defaults to the result of running config.guess. 
--host=host-type
the type of system on which the package runs. By default it is the same as the build machine. Specifying it enables the cross-compilation mode. 
--target=target-type
the type of system for which any compiler tools in the package produce code (rarely needed). By default, it is the same as host.

So if you use Linux to compile a GCC that should run under windows but targeting ARM, then your build type is linux, your host is windows, and your target is ARM.
For icarus, build is the system on which we compile it, and host is the system on which it runs. In our case, target is irrelevant, yet it is possible to set it via ./configure for some reason (maybe all autoconf-based project allow this regardless if this is necessary or not).

Anyhow, according to these definitions, dosify.exe, version.exe and draw_tt.exe should run on the build system.

steveicarus:
OK, I see your point, so lets change it to @BUILDCC@; but lets not change it to @CC@ because that is even more wrong.

udif:
OK, now I realize that CC does not come from autoconf but from the default 'make' value which is simply 'cc'. Since Ubuntu has a /usr/bin/cc on the path, this worked by pure luck.

I will take the time to look at the autoconf docs to see how I can get the correct C compiler for the build target.
In the meantime I also made sure today that the current patch can cleanly compile both 64 and 32 bit Windows versions:
On Windows 10 WSL (Ubuntu 14.04 based):

```
% apt-get install make flex bison gperf autoconf g++ mingw-w64
% git clone ....
% cd iverilog && autoconf
# For 64 bit builds
./configure --host=x86_64-w64-mingw32 && make
# For 32 bit builds
./configure --host=i686-w64-mingw32 && make
# For native builds
./configure

```

caryr:
If you change this from HOSTCC to something else make sure to change the appropriate FLAGS variable to match.

FYI I ran into this same problem a few years ago and never had time to figure out how to get these helper executables to get compiled correctly on the build machine. On one of my machines I may have a few other changes, but I think I pushed everything I had that made sense and left this for someone else. There is still an open bug report on SourceForge (https://sourceforge.net/p/iverilog/bugs/872/). If possible please verify that your change also fixes that issue and close the SF bug.

Also ideally this patch should go into both development and V10

udif:
After reading a bit (a lot actually) I think I have a better approach.
After finding this:
https://sourceware.org/autobook/autobook/autobook_143.html#Supporting-Cross-Compiler-in-Make
And then serching for CC_FOR_BUILD, I came across this:
http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_prog_cc_for_build.m4
After putting it in ac_local.m4, everything was straightforward.
On the way I updated the README file with cross-compile information.
@caryr , I haven't solved the CFLAGS problem, definitely not in any generic way. In our example only 3 files are compiled, and the utility programs apparently aren't affected by those flags in any important way.

udif:
Just added a small fix to the install target so that a few windows DLLs are copied from mingw with to the iverilog install target directory.
Without those DLLs, the resulting vvp.exe will not run.
Since this is a cross-compile, it is unreasonable to expect the users to already have those DLLs or to have them copy it from a Linux installation.

martinwhitaker:
I don't have time to check this at the moment. Have you tested that this doesn't affect the native WIndows/MSYS2 build/install?

udif:
Frankly, I haven't. The main reason I wrote this pull request was to save
me from building under Windows.
The problem is that at the moment there are no recen t Windows builds.
While Linux builds can be done with a simple sequence of 2-3 commands
(apt-get install / autoconf/config/make/make install),
a Windows build is more complex and requires much more manual intervention
(downloading, installing various tools, etc.).

udif:
I have taken the trouble to install MSYS2 and test this. The compilation instructions on the Wiki page installs WAY TOO MUCH stuff that's not needed, but lets ignor ethis atthe moment.
Since the current MSYS2 repository installs gcc 6.2.0, there are some new warnings, but the code compiles fine, and 'make check' works OK.
I added another patch for configure.in to disable those warnings. You may pick it too if you want.

steveicarus:
Should have merged this earlier. As it stands, I had to make a minor tweak to get it to merge now, since some files drifted a bit. So this could use some testing.

martinwhitaker:
I've tested a native Windows 10 build. 'make install' fails on the attempt to copy the additional DLLs. There's two reasons for that:

1. When using MSYS2, the DLLs are not in gcc's library search path.
2. If the user doesn't specify an install prefix, the default is to install to /mingw{32|64}/bin, which is where the DLLs are.

So if you want to keep that, we need to modify the Makefile.
BTW, the driver-vpi Makefile doesn't seem the right place for that.

PR #151, opened 2017-03-21
Added an option for ignoring errors about missing modules

tumik:
This commit adds an option (-i) to ignore errors about missing modules.

Ignoring missing modules is really useful for using iverilog as linter in text editors (like Atom or Sublime Text), where you usually only want to check single files without caring about modules on other files.

### DIFF
--- a/compiler.h
+++ b/compiler.h
@@ -117,6 +117,9 @@ extern bool debug_emit;
 extern bool debug_synth2;
 extern bool debug_optimizer;
 
+/* Ignore errors about missing modules */
+extern bool ignore_missing_modules;
+
 /* Control evaluation of functions at compile time:
  *   0 = only for functions in constant expressions
  *   1 = only for automatic functions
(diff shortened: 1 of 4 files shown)


steveicarus:
This looks fine and useful.

I am curious what happens when you use this flag in a non-linting situation.

jersey99:
Does seem really useful, and I gave this a try. However, if the missing file is included with an `include directive, ivlpp still errors out:

`Include file xxx.vh not found`

Not sure what the intended behavior in this case should be.

jfcherng:
@martinwhitaker I wonder can we have to PR also include in the `v10-branch`?

martinwhitaker:
We don't normally add new features to the stable branch, but this seems safe enough.
@steveicarus , what do you think?

jfcherng:
@martinwhitaker @steveicarus sorry to re-ping on this but any chance to land it on v10?

PR #181, opened 2017-12-16
vvp grammar: .port_info statements should require/generate semicolons.

### LINKED ISSUE #179: Bug in VVP grammar? .port_info statements don't end in a semicolon.
I discovered this while writing my own toy parser for vvp files.

Every single statement must end with a semicolon. with the small exception of `.port_info` statements, which a defined in `vvp/parse.y` (line 771) as not ending in a semicolon. I'm guessing this was an issue with the initial implementation of the port info feature. I've checked and there are no other exceptions.

Likewise, tgt-vvp skips the semicolon when outputting `.port_info` statements. 

It looks like it's reasonably easy to fix, but there is a question of backwards/forwards compatibility. My quick testing shows that adding a semicolon to the end of the `.port_info` statement will parse in current versions of vvp (I suspect it's using the empty statement rule). This means we can just update tgt-vvp to include the semicolon.    
But if we update the grammar in `vvp/parse.y` to enforce semicolons then old vvp scripts will no-longer work in newer versions. It might be a good idea to make the semicolon optional, and maybe print a warning message. Or we can just add a comment to `vvp/parse.y` explaining why it's missing for future reference.  

I'm more than happy to submit a pull request. I just wanted to consult on the correct way of dealing with backwards compatibility. 

Thanks.

phire:
This PR attempts to fix #179 

There are two parts. The first commit simply fixes tgt-vvp so it includes a semicolon after all .port_info statements. This fix is inherently backwards compatible, as current versions of vvp accept the extra semicolon without issue (they think it's an empty statement).

The second commit fixes the parser to be more strict and syntax error when the semicolon is missing. Because this would cause backwards compatibility issues (any vvp files generated on 11.0 dev before this commit is merged will be missing the semicolons), I've delayed the activation of this new stricter rule until 12.0

As vvp refuses to read files from older major versions, by the time work starts on 12.0, vvp will only be expected to parse vvp files generated with the correct semicolon syntax. At some point after 12.0, the second rule can be removed altogether.

### DIFF
--- a/tgt-vvp/vvp_scope.c
+++ b/tgt-vvp/vvp_scope.c
@@ -2304,7 +2304,7 @@ int draw_scope(ivl_scope_t net, ivl_scope_t parent)
             unsigned width = ivl_scope_mod_module_port_width(net,idx);
             if( name == 0 )
                 name = "";
-            fprintf( vvp_out, "    .port_info %u %s %u \"%s\"\n",
+            fprintf( vvp_out, "    .port_info %u %s %u \"%s\";\n",
                     idx, vvp_port_info_type_str(ptype), width,
 		    vvp_mangle_name(name) );
         }
--- a/vvp/parse.y
+++ b/vvp/parse.y
@@ -27,6 +27,7 @@
 # include  <cstdlib>
 # include  <cassert>
 # include  "ivl_alloc.h"
+# include  "version_base.h"
 
 /*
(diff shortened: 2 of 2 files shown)


martinwhitaker:
Adding the extra rule to the vvp parser introduced a shift/reduce conflict. I think we should just delete the old rule now.

PR #213, opened 2018-10-27
VPI search path updates

alexforencich:
This patch does three main things. First, it reverses the order in which the internal path list is checked so that paths specified with -M are checked first and the default paths are checked last.  Second, it enforces the path count limit to prevent crashes.  Third, it implements the environment variable IVERILOG_VPI_MODULE_PATH to specify a colon-separated list of directories in which to look for VPI modules.

### DIFF
--- a/vvp/main.cc
+++ b/vvp/main.cc
@@ -262,43 +262,6 @@ int main(int argc, char*argv[])
       extern bool stop_is_finish;
       extern int  stop_is_finish_exit_code;
 
-#ifdef __MINGW32__
-	/* Calculate the module path from the path to the command.
-	   This is necessary because of the installation process on
-	   Windows. Mostly, it is those darn drive letters, but oh
-	   well. We know the command path is formed like this:
-
-		D:\iverilog\bin\iverilog.exe
-
-	   The IVL_ROOT in a Windows installation is the path:
-
-		D:\iverilog\lib\ivl$(suffix)
-
-	   so we chop the file name and the last directory by
-	   turning the last two \ characters to null. Then we append
-	   the lib\ivl$(suffix) to finish. */
-      char *s;
-      char basepath[4096], tmp[4096];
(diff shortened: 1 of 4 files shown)


martinwhitaker:
I'm uncomfortable about reversing the order of search paths supplied on the command line - it's not what users will expect (and who reads the man page!). I do agree they should take precedence over the environment variable and the default path.

alexforencich:
I think it's sort of the default behavior for most applications - the arguments are processed from left to right in order, and each one inserts a new 'highest priority' path that supersedes the previous ones, resulting in the last argument taking precedence over the rest.  Also, if the order is not reversed, then either the command line arguments need to be parsed in reverse order, or the list needs to be built up by pushing items in at index 0 and moving any other items down the list to make room.

martinwhitaker:
It's not the default for gcc, either for -I or for -L.

If the order hadn't been specified before, I wouldn't be so bothered. But it was, so some users may have scripts that rely on it. We shouldn't change existing behaviour without a good reason - and making it easier to implement is not a good enough reason.

martinwhitaker:
If you process the environment variable after the command line arguments, then tack the default path on at the end, everything will end up in the right order.

alexforencich:
Alright, I will rewrite all of that, then.

alexforencich:
Updated.  You guys really need to do a tabs to spaces swap.  Using tabs for indents is fine if the indent width is one tab width, but you guys are indenting by 6 spaces with a tab width of 8.

martinwhitaker:
OK, looks good apart from the '-M -' option having changed behaviour - it used to remove the default path. I've fixed that, although it's probably not needed now we put the default path last. I also took the opportunity to refactor the code for easier readability/maintainability.

As for indentation, if it was up to me, the hard tabs would have gone long ago. But it's Steve's project, so that's up to him.

Issue #248, opened 2019-06-04
Support for attributes according to Verilog 2005 standard


mkurc-ant:
It seems that the Icarus Verilog is unable to parse some cases of attribute specification which are allowed by the Verilog 2005 standard. 

I prepared some failing cases to demonstrate the issue. For all of them the Icarus Verilog was invoked with the command `iverilog -g2005 <verilog_file>`:

- Attributes on port connections:

Input Verilog code (attrib05_port_conn.v):
```
module bar(clk, rst, inp, out);
  input  wire clk;
  input  wire rst;
  input  wire inp;
  output reg  out;

  always @(posedge clk)
    if (rst) out <= 1'd0;
    else     out <= ~inp;

endmodule

module foo(clk, rst, inp, out);
  input  wire clk;
  input  wire rst;
  input  wire inp;
  output wire out;

  bar bar_instance ( (* clock_connected *) clk, rst, (* this_is_the_input *) inp, out);
endmodule
```

Error message:
```
attrib05_port_conn.v:19: syntax error
attrib05_port_conn.v:19: error: Syntax error in instance port expression(s).
```

- Attributes on function calls (attrib07_func_call.v):

Input Verilog code:
```
module foo(clk, rst, inp_a, inp_b, out);
  input  wire clk;
  input  wire rst;
  input  wire [7:0] inp_a;
  input  wire [7:0] inp_b;
  output wire [7:0] out;

  function [7:0] do_add;
    input [7:0] inp_a;
    input [7:0] inp_b;

    do_add = inp_a + inp_b;
  endfunction

  always @(posedge clk)
    if (rst) out <= 0;
    else     out <= do_add (* combinational_adder *) (inp_a, inp_b);

endmodule
```

Error message:
```
attrib07_func_call.v:19: syntax error
attrib07_func_call.v:19: error: malformed statement
```

martinwhitaker:
Pull request now merged. Thanks.

PR #249, opened 2019-06-06
Support for attributes on port connections and function calls

mkurc-ant:
This PR adds support for the Verilog parser so it can accept attribute specifications on module instance port connections and on function calls. Solves issue https://github.com/steveicarus/iverilog/issues/248

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -597,6 +597,7 @@ static void current_function_set_statement(const YYLTYPE&loc, vector<Statement*>
 
 %type <named_pexpr> modport_simple_port port_name parameter_value_byname
 %type <named_pexprs> port_name_list parameter_value_byname_list
+%type <exprs> port_conn_expression_list_with_nuls
 
 %type <named_pexpr> attribute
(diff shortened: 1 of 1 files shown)


mithro:
@mkurc-ant Can you add tests for this?

mkurc-ant:
@mithro Tests added. They can be invoked by running `make test` or during execution of `make check`

martinwhitaker:
Tests should be added to the test suite, not to the compiler source tree. See https://iverilog.fandom.com/wiki/Developer_Guide#Testing_Your_Work.

mkurc-ant:
@martinwhitaker As you suggested I removed tests from the source tree and added them to the test suite. I created the second PR for tests: https://github.com/steveicarus/ivtest/pull/6

martinwhitaker:
Sorry it took so long for someone to look at this.. Reviewing the changes, it seems they are a bit over-permissive - the standard only allows attributes on module port connections, not on gate primitive connections. But I don't think we'll worry about that - it does no harm! Thanks for the contribution.

I've also back-ported this to the v10 branch.

PR #250, opened 2019-06-07
Degrade primitive redefinition and ifnone with edge-sensitive-path to warnings

thasti:
While trying to simulate a larger ASIC design (together with some technology libraries), I came across two major annoyances:
  - The technology files redefine primitives with the same content in multiple places, and can not easily be altered
  - The technology files make heavy use of timing checks, many of them using the unsupported ifnone+edge-sensitive path construct.

In the end I got my simulations working and maybe the changes I needed to do to iverilog can be acceptable for mainline. As altering the vendor libraries is not really an option for me and these two things seem to be commonly used, maybe it can increase adoption of iverilog for such uses.

As it was at times hard for me to figure out, which messages from iverilog are errors and which are warnings, I also tried improving the distinction between the two.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -5752,7 +5752,7 @@ specify_item
 		  pform_module_specify_path(tmp);
 		}
 	| K_ifnone specify_edge_path_decl ';'
-		{ yyerror(@1, "Sorry: ifnone with an edge-sensitive path is "
+		{ yywarn(@1, "Sorry: ifnone with an edge-sensitive path is "
 		              "not supported.");
 		  yyerrok;
 		}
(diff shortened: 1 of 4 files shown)


martinwhitaker:
There are calls to yyerror/VLerror that already include the "error:", so we don't want to add it a second time. And other calls that use "sorry:" instead of "error:". So I think you need to remove that change.

thasti:
Good catch, I removed it. Thanks for the feedback!

martinwhitaker:
Seems OK to me. I won't backport this to v10, as it is a change in behaviour but not a bug as such.

thasti:
Fine with me - thank you for merging. Happy to see this in the next release then :)

PR #267, opened 2019-09-29
replace deprecated yacc directives

sielicki:
bison 3.4.2 complains when encountering '%pure-parser':
"warning: deprecated directive, use ‘%define api.pure’"

A quick google suggests this option has been around since at least
2012, maybe longer, so probably safe to replace.

Signed-off-by: Nicholas Sielicki <sielicki@yandex.com>

### DIFF
--- a/tgt-pcb/fp.y
+++ b/tgt-pcb/fp.y
@@ -1,5 +1,5 @@
 
-%pure-parser
+%define api.pure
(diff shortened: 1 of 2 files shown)


steveicarus:
Seems harmless enough. Merged.

sielicki:
Thanks for the quick response and merge!

PR #269, opened 2019-10-06
Validate that value is  handled for case priority and unique qualifiers

nbrereton:
Simulator should warn if a case statement is qualified with priorty or unique and the value is not handled.

### DIFF
--- a/tgt-vvp/vvp_process.c
+++ b/tgt-vvp/vvp_process.c
@@ -652,9 +652,9 @@ static int show_stmt_case(ivl_statement_t net, ivl_scope_t sscope)
 
       unsigned idx, default_case;
 
-      if (qual != IVL_CASE_QUALITY_BASIC) {
+      if (qual != IVL_CASE_QUALITY_BASIC && qual != IVL_CASE_QUALITY_PRIORITY) {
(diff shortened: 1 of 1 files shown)


steveicarus:
Should a priority case statement issue a warning on the default case, even if the default case is explicitly specified? Your patch suggests you believe no. We'll leave it at that for now as I don't know the correct answer. It's tempting to put in a compile time warning (during elaboration) if defaults are used in a priority case statement, to warn about the decision we made here.

steveicarus:
We could use a test of this to go into the ivtest regression test suite.

nbrereton:
RIght, I don't believe that a warning should be issued if a default case is specified.

Looking at http://ecee.colorado.edu/~mathys/ecen2350/IntelSoftware/pdf/IEEE_Std1800-2017_8299595.pdf section 12.5.3 "the simulator shall issue a violation report if no case_item matches", I believe that an explicit default case would fall into that; also section 12.4.2 is a bit more explicit, saying that "a violation report shall be issued if no condition matches unless there is an explicit else." which I think would by induction also cover the default case case (!).

I'll add a test for ivtest shortly.

PR #66, opened 2015-05-05
Add travis-ci integration (linux + windows)

themperek:
This will add a simple configuration file for free continuous integration service https://travis-ci.org.
It will build and run the ivtest on every commit on linux and windows . It integrates with github and also builds all pull requests giving a status.
Example build log: https://travis-ci.org/themperek/iverilog

It has to be enabled by @steveicarus see:  http://docs.travis-ci.com/user/getting-started/#Step-one%3A-Sign-in 

To get proper status for ivtest one need to add exit 0/1 to vvp_reg.pl so travis knows if it passes or not. 

Suggestion:
It is maybe easier to keep it clean when ivtest is part of iverilog repository.

Best Regards,
Tomasz

PR #275, opened 2019-10-16
Fix MinGW compilation issues

sverlan:
The current version does not compile in mingw (using steps described on https://iverilog.fandom.com/wiki/Installation_using_MSYS2). 

A quick review of the code had shown that this is due to recently introduced changes. There are two problems that break the compilation (using gcc-8.3.0):
-  ivl_stmt_case_quality is not defined in ivl.def so tgt-stub cannot be linked
- mmap.h is absent in msys2, so MAP_FAILED is not defined (breaking the compilation of vpi/fstapi.c)

 By making the corresponding changes it compiles perfectly.

### DIFF
--- a/ivl.def
+++ b/ivl.def
@@ -302,6 +302,7 @@ ivl_stmt_parm_count
 ivl_stmt_rval
 ivl_stmt_sfunc_as_task
 ivl_stmt_sub_stmt
+ivl_stmt_case_quality
 
 ivl_switch_a
 ivl_switch_b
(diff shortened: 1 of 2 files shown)


caryr:
We should not update fstapi.c since this is copied directly from GTKWave. I will forward this pull request to Tony so he can fix this upstream.

themperek:
BTW #66 would help discover such issues early.

sverlan:
This can be also dirty fixed by a define in the makefile...
BTW, I discovered that make install does not strip the files, so they are really huge...

caryr:
This should be fixed in the latest development and V10 head. Tony updated GTKWave and I just pushed those changes to Icarus. I also added the missing line in ivl.def. Since your patch did not matching our coding standard I made the change myself.

If this solves your issues please close this request.

caryr:
That can be done by overriding LDFLAGS during configuration (e.g. env LDFLAGS=-s ./configure ...). I believe Steve intended the default build to include debugging flags, though it could easily be argued that the normal user would prefer the smallest and most optimized version and those that need debug capabilities know enough to tweak the compilation flags.

sverlan:
ok, thank you for pointing this out. I'm closing the issue.

PR #282, opened 2019-11-12
Update lexor.lex

getziadz:
Change on the ivlpp lex file to always append a newline character after EOF. This solves the problem when concatenating files using -E with files that do not finish with a newline character.

Example results can be in the form 'endmodulemodule <module_name>(...)'.

This change will result to the following:
'endmodule
module <module_name>(...)'

### DIFF
--- a/ivlpp/lexor.lex
+++ b/ivlpp/lexor.lex
@@ -625,7 +625,7 @@ keywords (include|define|undef|ifdef|ifndef|else|elseif|endif)
     fputc('\n', yyout);
 }
 
-<<EOF>> { if (!load_next_input()) yyterminate(); }
+<<EOF>> { fputc('\n', yyout); if (!load_next_input()) yyterminate(); }
 
 %%
  /* Defined macros are kept in this table for convenient lookup. As


Jbalkind:
Also experiencing this issue - it's hidden when not using `-E` because `ivlpp` is called with `-L` which adds line declarations at the start of each file.

steveicarus:
Can you change this so that it only appends a \n if the output doesn't already end with a newline? I'm concerned with line counts becoming off.

getziadz:
Hi Steve,
   Sorry for the delay of this update. I've come with something up that I need to test more.
However, the solution assumes disabling 'nounput' and 'noinput'. Is that OK with you or do I need to find a way around that?

steveicarus:
I think it is OK to use the flex generated unput() and input(). I don't think we are doing anything tricky that would make them fail or make a mess.

Whatever you do, I would like you to try to test to make sure that line numbers after includes remain correct. The ivlpp preprocess generates `line directives, and the ivl compiler reads those to update the line counters so that error messages give the correct file/line information.

getziadz:
Sounds good. I'll try to finish this ASAP. Is there some specific testbench I should be using to verifying the line numbers or something custom is fine?

steveicarus:
There may be tests in the ivtest project for this, so you should at the very least run your changes through that. If you can come up with tests, that would be very helpful.

getziadz:
OK. I'll take a look at the ivtest project and if I think something more is needed I will included it.

getziadz:
This may take longer than expected. I have attempted to create a solution in two different ways but it leads to failures on some ivtest benchmarks. As I don't have many cycles for this work it may take a while to have something working. If you can't wait on that we can create an issue on github so that someone else could try and submit a solution.

steveicarus:
This PR has gone silent, and is obsolete. Closing.

PR #308, opened 2020-02-05
Fix bad bounds check in for loop

blackgnezdo:


martinwhitaker:
Applied to both master and v10 branches. Regression test added to the test suite. Thanks for the fix.

PR #311, opened 2020-02-15
Replace assertion in evaluate_index_prefix with useful error message

nmoroze:
Currently, if Icarus encounters chained range selections during compilation, i.e.: `assign b = a[4:0][1:0];`, it fails an assertion and crashes. I've changed this behavior to instead have a useful error message, so that the user knows what went wrong and can fix their code.


martinwhitaker:
Looks good to me. Thanks for the contribution.

PR #328, opened 2020-05-19
Allow +libext+ and -y commands to be given in any order (and small fix in manpage)

purdeaandrei:


martinwhitaker:
std::tuple is a C++11 feature, so this will fail when building with older compilers. You should use std::pair instead.

Also, please conform to the 6 space indentation used in the Icarus source code.

purdeaandrei:
@martinwhitaker 
> std::tuple is a C++11 feature, so this will fail when building with older compilers. You should use std::pair instead.

Alright, I changed to pair, and tested it, please see my latest commit.

> Also, please conform to the 6 space indentation used in the Icarus source code.

Could you clarify where it is that I'm not following indentation used by existing code?
As far as I can see, it was already okay.
I mean the source code is using a mix of tab characters and space characters, and I
assume that the convention is 6 spaces indentation, which can be composed out of
tabs and/or spaces, as long as tabs are interpreted to be 8 spaces long.

The code is following exactly this scheme as far as I can see.

The body of the for loop is 6 'spaces' indented compared to where the "for" keyword sits,
so the total indentation is 12 spaces, made out of 1 8-space tab and 4 real spaces.
however the line after the "for" loop is only 5 spaces deeper then the for, because it's not inside the body, it is instead aligned to the first item inside the parenthesis of the for loop. This is 
exactly how other code looks in the file, for example there is lower down another long if
statement which has the same type of alignment.

purdeaandrei:
I think the issue is that github is showing the changeset incorrectly. For whatever reason github decided to use 8-space tabs on the top hunks of my changeset, and 6-space tabs on the bottom hunk. I think github is trying to be smart about how long to interpret the tabs to be, but it's applying these smarts to each hunk individually, so it ends up showing weird indentation.

Please check out my changeset in your favorite editor that shows visible tabs, to see that it's correct.

martinwhitaker:
You're right, GitHub is displaying the patch wrong. Looks fine in my text editor.
Your latest patch is failing on Travis CI.

purdeaandrei:
> Your latest patch is failing on Travis CI.

Okay, i now compiled with gcc5 and now I see the issues on my machine too.
Fix pushed...

martinwhitaker:
Looks good now. Thanks for your contribution.

PR #335, opened 2020-06-23
Fix segmentation fault when returning value from void function

purdeaandrei:


purdeaandrei:
Test added in one of the commits from here: https://github.com/steveicarus/ivtest/pull/16

martinwhitaker:
I've committed a simpler fix for this issue.

purdeaandrei:
@martinwhitaker 
Indeed, yours is a much simpler overall solution, however my gut is telling me, in addition to that to add an assert that res is not null, after the first line of this snippet:
```
NetNet*res = target->find_signal(target->basename());
ivl_variable_type_t lv_type = res->data_type();
```
This function can sometimes return NULL, and it may be true that at this point we know that we are not elaborating a void function, therefore it can't be NULL, but it's not obvious from this snippet.

Do you agree?

PR #347, opened 2020-07-25
aclocal.m4: fix issue in cross-compiling

vowstar:
The aclocal.m4 called nm directly.

It can cause issue in cross-compiling and because is not possible use a different NM implementation (like llvm-nm).

The compile error log:
https://bugs.gentoo.org/attachment.cgi?id=648556

Closes: https://bugs.gentoo.org/731906

### DIFF
--- a/aclocal.m4
+++ b/aclocal.m4
@@ -23,10 +23,10 @@ AC_SUBST(install_suffix)
 # ------------------------------
 # Sub-macro for AX_C_UNDERSCORES_LEADING and AX_C_UNDERSCORES_TRAILING.
 # Unwarranted assumptions:
-#   - the object file produced by AC_COMPILE_IFELSE is called "conftest.$ac_objext"
-#   - the nm(1) utility is available, and its name is "nm".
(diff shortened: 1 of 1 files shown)


aclocal.m4
```diff
@@ -25,9 +25,11 @@ AC_SUBST(install_suffix)
 # Unwarranted assumptions:
 #   - the object file produced by AC_COMPILE_IFELSE is called "conftest.$ac_objext"
 #   - the nm(1) utility is available, and its name is "nm".
+# Here use $NM because they can cause issue in cross-compiling and because is
+# not possible use a different NM implementation (like llvm-nm)
```
martinwhitaker, line 29: This comment contradicts the previous comment line. I would suggest something like:
#   - the nm(1) utility or an equivalent is available, and its name is defined by the $NM variable.

vowstar: Yes, this will looks better, the comment merged

martinwhitaker:
Looks good to me.

PR #349, opened 2020-07-29
Bison includes its generated header in *.cc. Generate with correct name.

hzeller:
The current bison (3.7) generates a *.cc file that includes the header
it generated. For parse.cc this would be parse.hh. Right now, we rename
this header to have a common name used in other files, but this results
in a compile error for the parse.cc file:

parse.cc:462:10: fatal error: parse.hh: No such file or directory
  462 | #include "parse.hh"
      |          ^~~~~~~~~~

Fix this by telling bison to output the header file to the correct
filename in the first place so that we don't have to rename it.
(using the --defines instead of -d option).

This looks like a bison specific option not available in Posix yacc;
but looks like we're requiring bison anyway.

Signed-off-by: Henner Zeller <h.zeller@acm.org>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -256,10 +256,8 @@ parse.o: parse.cc
 
 # Build this in two steps to avoid parallel build issues (see pr3462585)
 parse.cc: $(srcdir)/parse.y
-	$(YACC) --verbose -t -p VL -d -o $@ $<
+	$(YACC) --verbose -t -p VL --defines=parse.h -o $@ $<
 parse.h: parse.cc
-	mv parse.cc.h $@ 2>/dev/null || mv parse.hh $@
-	touch $@
 
 syn-rules.cc: $(srcdir)/syn-rules.y
 	$(YACC) --verbose -t -p syn_ -o $@ $<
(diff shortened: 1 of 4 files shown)


hzeller:
Confirmed that the `--defines` option in bison is there since at least 2000 (first mention in the bison Changelog file is at `2000-10-02` about testing that option, so its support is likely there before).
So this change is compatible with any relevant bison version iverilog would be built.

martinwhitaker:
I've tested this works with bison 3.01, and we specify bison 3+ in the README.txt file, so that's good enough.
The "parse.h: parse.cc" rules don't achieve anything now, but I'll deal with that in a separate commit.

hzeller:
The header rules don't _do_ anything, but they are still important to express the dependency graph: some targets only depend on the header file and the Makefile needs to know where they come from.

So the `parse.h: parse.cc` rules still need to be there even if there is no action associated.

Apparently it is not possible to just have a rule where both resulting targets are on the left (as in `parse.h parse.cc : parse.y`), as this broke parallel building (`make -j`) back in the day; so it was explicitly split in two rules in this change from 2012-01-04 https://github.com/steveicarus/iverilog/commit/9c99bce054c67c59435f2ef0e5dfc6698bebe2dc

martinwhitaker:
But the header rules don't help - if you delete one of the header files but leave the .cc file intact, the next build will fail.
See my recent commit for how this can be fixed.

hzeller:
Ah, I see what you mean.

PR #354, opened 2020-08-05
Added Dockerfile that may be used for CI/CD tasks

Thirsty2:
Added dockerfiles directory, and an Alpine Linux based Dockerfile that builds iverilog in a container, then copies the binary to a new container in a second stage to reduce the size of the iverilog container to less than 90 MBytes.  There are instructions in a pair of readme.txt files, and comments in the Dockerfile to demonstrate how to build and run the container.


Thirsty2:
I should probably start over, and do this in a branch, or simply delete this repo and re-fork.  I did the work on master, merging from upstream as I went, and pull request tests are failing.

PR #357, opened 2020-08-07
Add Dockerfile that builds iverilog in an alpine linux container for …

Thirsty2:
…CI/CD use.

### DIFF
--- a/dockerfiles/alpine312/Dockerfile
+++ b/dockerfiles/alpine312/Dockerfile
@@ -1,0 +1,68 @@
+FROM alpine:3.12 as builder
+
+# The following will be cached by the docker layer caching mechanism the first time this
+# dockerfile is built, and these commands will never execute again unless you force them
+# to with the --no-cache option:
+# docker build --no-cache . -t iverilog:latest
+RUN apk add --no-cache \
+	autoconf \
+	build-base \
+	bison \
+	bzip2-dev \
+	flex \
+	git \
+	gperf \
+	readline-dev \
+	zlib-dev
+
+# This clone command results in a bare repo that will remain at the state it
+# is at when this Dockerfile is first built and cached by docker. 
+# This leverages the caching to avoid repeated full clones, saving bandwidth.
+RUN git clone --mirror https://github.com/steveicarus/iverilog iverilog.git
+
+# Bust the cache here or the following 'git remote update' will not execute.
+# Pass a random value to the BUST_THE_CAST ENV declared below for 
+# every build where you want to bust the cache.  As an alternative to a 
+# random variable, you could use a version string or git hash or similar
+# to bust the cache only when your (branch, source, whatever) is updated.
+# The following build command is a good way to bust the cache here.
+# docker build --build-arg IVERILOG_BRANCH=master --build-arg BUST_THE_CACHE=$(date +%s) . -t iverilog:latest
+ARG BUST_THE_CACHE="cache not busted"
+RUN echo "${BUST_THE_CACHE}" > bust_the_cache.txt
+RUN cd iverilog.git && \
+    git remote update && \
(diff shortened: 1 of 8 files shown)


martinwhitaker:
I'm leaving it to Steve to decide whether he wants to add this.

My only comment is why not use one of the tests in the examples directory (as used by `make check`) rather than creating a new one.

Thirsty2:
Thanks Martin, Steve.

The goal of this container is to allow automation of iverilog simulations
in automated build systems.

The dockerfiles/test directory probably should have been named example or
demo instead of test.  It doesn't actually check the results of the test,
but rather provides a simple example of how to create a Makefile that the
docker container will use to compile the .v files and run vvp to create a
.vcd output.  This particular Dockerfile uses make as its entrypoint, which
means you can't easily run commands other than make from the container it
produces.  In fact, you can't actually run make in the container yourself -
it is run automatically when the container starts.  You have control over
the Makefile.  You have the ability to pass parameters to make (the
Makefile contains a clean: section, and the readme.txt file describes how
to add clean at the end of the docker run command.)

While this is limiting in many ways, it demonstrates a simple way to
incorporate the iverilog container into an automated build/test system.
This allows you to run iverilog on something like a Jenkins server or a
github action in the cloud without installing it first.  You don't have to
install make either, as this is included and run from within the container.

If you modify the Dockerfile to remove the entrypoint, and rebuild it, you
can execute an interactive bash shell and do a lot more, but limiting what
can be done is useful on automated build systems.  Perhaps there is
something better to use as an entrypoint than make, but most everyone
understands make so it is a reasonable demonstration.

Containers are ideal for automated builds because they are immutable, and
guaranteed to start with exactly the same initial state every time.  You
don't have to worry about someone tweaking a configuration setting to make
something work and forgetting to document it or commit it.  Everything that
is done is visible in the Dockerfile, which is easily version controlled.
Every time you run the container (with the docker run command) you get a
brand new virgin instance of it, identical to the last one.  Because of
these properties, they are preferred even over VMs for automating builds.

Doug

Thirsty2:
The container built by the Dockerfile's final stage does not contain gcc/g++.  It is a minimal container stripped of everything except the iverilog binaries and the make utility, which it runs automatically at startup (the entrypoint at the end of the Dockerfile makes sure this container always runs make when it starts.)  The size of the image is roughly 87.8 MB for the latest github commit.  It is designed to run iverilog as part of a CI/DC build system.

However, there is an intermediate container (in this case it is the first stage) that does contain the gcc/g++ tools, and can run the 'make check' command as part of the build.  You can actually run this container as well, though it is not tagged by default.  Use the 'docker image ls' command to find it.  It will be the container built just before the final tagged container.  The size is roughly 624 MB for the first stage 'builder' container, and 87.8 MB for the final stage iverilog:latest container.

If you prefer to build and tag this intermediate container, you can do so with a command like:
docker build --target builder . -t iverilog-builder

The above command stops after building the first stage, and the result is a 624 MB image tagged iverilog-builder:latest.

Then you can run it with a command like:
docker run -it iverilog-builder:latest sh

This runs an interactive sh inside the container without mapping any directories from the host.  Every file you see inside the container is in the container's image, not mapped from the host.

You can cd into iverilog and run make commands such as 'make check', 'make clean' etc.

Modifying files inside this container won't have any effect on the host filesystem, and the next time you run the container it gets reset to its image's initial state with no changes.

If you do want to persist changes to your filesystem, a good way to do this is to use the -v command line parameter when running the container.  There are examples of this in the provided readme.txt files.

While this is fun for poking around, and useful in its own right as a tool to build iverilog in a container, it is much larger than the 87.8MB output image of the final stage, and contains more than it should for simply running iverilog as part of a CI/CD build of verilog files.  

Doug

PR #371, opened 2020-09-20
parse.y: Mark enum output ports always as IMPLICT_REG

larsclausen:
Unless explicitly declared a wire an enum output port is of variable type
and should be marked as IMPLICT_REG.

Currently this is only done when the base type of the enum is `logic`. But
it should be done for all enums regardless of their base type.

Without this change for example the following snippet

```systemverilog
  typedef enum {
      A,
      B
  } E;

  module M (
      input E ei,
      output E eo
  );

      always_comb eo = ei;

  endmodule
```

fails with the following error message
```
  test_enum.sv:11: error: eo is not a valid l-value in M.
  test_enum.sv:8:       : eo is declared here as wire.
```

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -4659,9 +4659,8 @@ port_declaration
 		    use_type = NetNet::IMPLICIT_REG;
 	      } else if (dynamic_cast<struct_type_t*> ($4)) {
 		    use_type = NetNet::IMPLICIT_REG;
-	      } else if (enum_type_t*etype = dynamic_cast<enum_type_t*> ($4)) {
-		    if(etype->base_type == IVL_VT_LOGIC)
-			use_type = NetNet::IMPLICIT_REG;
+	      } else if (dynamic_cast<enum_type_t*> ($4)) {
+		    use_type = NetNet::IMPLICIT_REG;
 	      }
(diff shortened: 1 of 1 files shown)


steveicarus:
Can you attach a test program that we can include in the ivtest project (https://github.com/steveicarus/ivtest)?

Also, it's been a little while since I've been knee-deep in the LRM. Can you point me to some references that justify this? (I don't doubt you. It's just for my edification.)

larsclausen:
There is nothing in the LRM that would justify the current behavior. There is no mention of any base type specific behavior. 

`23.2.2.3 Rules for determining port kind, data type, and direction` says
```
For output ports, the default port kind depends on how the data type is specified:
- If the data type is omitted or declared with the implicit_data_type syntax, the port kind
shall default to a net of default net type.
- If the data type is declared with the explicit data_type syntax, the port kind shall default to
variable.
```

An enum falls into the second category.

You could also argue that it has to be this way because `output reg E e` is invalid syntax, where as `output wire E e` is valid. So it must be variable type unless explicitly declared as wire type, because otherwise there would be no way to get a variable type enum port.

PR #373, opened 2020-10-04
Dockerfile with GitHub actions

Thirsty2:
This Dockerfile and associated github actions builds an image based on Alpine Linux 3.12.  The docker build executes regression tests automatically, and publishes the container image as a github package if the build and tests succeed.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,76 @@
+name: Docker
+
+on:
+  push:
+    # Publish `master` as Docker `latest` image.
+    branches:
+      - master
+
+    # Publish `v1.2.3` tags as releases.
+    tags:
+      - v*
+
+  # Run tests for any PRs.
+  pull_request:
+
+env:
+  # TODO: Change variable to your image's name.
+  IMAGE_NAME: iverilog
+
+jobs:
+  # Run tests.
+  # See also https://docs.docker.com/docker-hub/builds/automated-testing/
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run tests
+        run: |
+          if [ -f docker-compose.test.yml ]; then
+            docker-compose --file docker-compose.test.yml build
+            docker-compose --file docker-compose.test.yml run sut
+          else
+            docker build --build-arg GITHUB_WORKSPACE=$GITHUB_WORKSPACE . --file Dockerfile
+          fi
+
+  # Push image to GitHub Packages.
+  # See also https://docs.docker.com/docker-hub/builds/
+  push:
+    # Ensure test job passes before pushing image.
+    needs: test
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build image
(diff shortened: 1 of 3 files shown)


caryr:
@steveicarus is there any reason this should not be merged. Do we need more information regarding exactly what this is doing/how it will be used? @martinwhitaker do you have any comments? This all seems reasonable, but I must admit I've not looked at Docker much.

caryr:
And a question for @Thirsty2 what is your intention for having a containerized version of Icarus? I can think of some, but would appreciate understanding your perspective as well.

Thirsty2:
The primary use case is for build automation.

Doug

caryr:
I assume this is build automation for you (e.g in an academic/learning setting someone would build a container and then that could be used by all students with no system, version, etc issues).

martinwhitaker:
I don't have any objections to this, but I don't understand the need to include it either.

caryr:
Yes Martin, that's basically where I am. I can see where this could be useful for the case I mentioned earlier and maybe there are others I've not thought of. I will admit my vision is likely clouded by years of building a huge number of tools from source so all this has mostly become second nature to me.

To me the biggest issue preventing me from saying this is a no brainier is you would like to have GTKWave available, but from what I read GUI programs are not really supported.

Thirsty2:
We could easily add GTKWave to the dockerfile, but I personally prefer to
keep them separate.  I have GTKWave running in a Docker container, and it
works very well.  I am happy to provide a dockerfile for that if you want
to try it.

As you mentioned, another great use case is students, or anyone that wants
to try iverilog without having to install it or compile it themselves.
Perhaps someone has a tried and true version of iverilog and they are
nervous about updating it.  Trying it under a Docker container allows them
to make sure nothing breaks for their use case before they take the plunge
of updating.  Likewise, they can try an older version that they knew worked
for them in a container if they are seeing some odd behavior after
updating.  Not that this ever happens with iverilog, but in general the
ability to run different versions of software with no worry about conflicts
is a nice thing for troubleshooting.

Another good reason to have a dockerfile is that it acts as an executable
version of documentation that makes the dependencies and the process to
build iverilog very plain.

Merging this change would be very low risk, as it doesn't affect the
iverilog code in any way.

Doug

steveicarus:
I think the Dockerfile files are OK as a source of build documentation, if nothing else. They are harmless.

However, I wonder if the deploy action would eat into our free action compute time budget. The CI actions are already time consuming. do we really want to include docker build and deploys with all the other build actions?

caryr:
I believe the deploy should certainly not be part of the normal CI. For stable it should only be for an actual release, for devel maybe quarterly or monthly or maybe never. Is there a different pool that can be used for the docker build as compared to our existing CI?

caryr:
Hi Doug, I assume this is just the recipe for building containers and there is no risk of this creating extra load where it builds every time we push a change. Once you confirm I will merge this.

Thirsty2:
As it is now, it would build and test the container with every push to
master.  See line 7 of the docker-publish.yml file.  If you merge this and
build it once, then remove line 7 (and you may as well remove 8 as well
since you won't have that branch) then it won't build with every push.  It
will still build with each tagged release (lines 11 and 12.)

It is not too difficult to add the ability for you (or someone with admin
access to the repo) to build on demand by pushing a button.  I can show you
how to do this if you like.

Doug

DeflateAwning:
This would be great to get integrated into the project!

PR #400, opened 2020-11-29
ci: migrate from Travis CI to GitHub Actions

umarcor:
In this PR, CI is migrated from Travis CI to a GitHub Actions workflow, as is. CI will run for any push or PR. Unlike Travis, GitHub Actions is free "without restrictions" for open source (public) projects. At the same time, it should be faster, since installing MSYS2 through chocolatey is avoided. 

Windows execution time:

- On Travis: 28min https://travis-ci.com/github/steveicarus/iverilog/builds/203958045
- On Windows: 31min https://github.com/umarcor/iverilog/runs/1469769023?check_suite_focus=true
- On Windows, without chocolatey: 20min https://github.com/umarcor/iverilog/runs/1469803219?check_suite_focus=true

PR #398, opened 2020-11-29
CI: test on MINGW64 and MINGW32, use PKGBUILD recipe

umarcor:
**This PR is not to be merged per se, but for showcasing a problem**.

In the context of [hdl/MINGW-packages](https://github.com/hdl/MINGW-packages), I'm writing a PKGBUILD file for having `iverilog` upstreamed as an MSYS2 package ([packages.msys2.org](https://packages.msys2.org/package/)). See [hdl/MINGW-packages@iverilog: mingw-w64-iverilog](https://github.com/hdl/MINGW-packages/tree/iverilog/mingw-w64-iverilog).

So far, `make`, `make check` and `make install` run successfully. Unfortunately, when testing the resulting package, I get `Segmentation fault` on MINGW64 (it passes on MINGW32).

This PR contains a PKGBUILD similar to the one in hdl/MINGW-packages, but here sources are not retrieved, since the build is executed in subdir `msys2` (new, where PKGBUILD is located). Furthermore, a GitHub Actions workflow is added. It uses Action [setup-msys2](https://github.com/msys2/setup-msys2), mimicking the workflow in the upstream [msys2/MINGW-packages](https://github.com/msys2/MINGW-packages).

- Execution of the workflow: https://github.com/umarcor/iverilog/actions/runs/389535247
- `make check` passing on MINGW64: https://github.com/umarcor/iverilog/runs/1469150810?check_suite_focus=true#step:5:1378
- `make check` passing on MINGW32: https://github.com/umarcor/iverilog/runs/1469150812?check_suite_focus=true#step:5:1380
- Simple test **failing** on MINGW64: https://github.com/umarcor/iverilog/runs/1469150810?check_suite_focus=true#step:7:15
- Simple test and ivtest passing on MINGW32: https://github.com/umarcor/iverilog/runs/1469150812?check_suite_focus=true#step:7:15

NOTE: the simple test is a random repo I found in GitHub.

Friendly nudge to some users who have been recently dealing with building and using iverilog on Windows: @martinwhitaker @ktbarrett @garmin-mjames

PR #402, opened 2020-11-30
ci/win: install MSYS2 packages through setup-msys2 (cached) 

umarcor:
As discussed in #400, this PR updates the Windows job in GitHub Actions, in order to use Action msys2/setup-msys2 for updating the built-in MSYS2 installation, and for adding dependencies. By default, the Action caches installed packages.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,16 +50,26 @@ jobs:
 
     - uses: actions/checkout@v2
 
+    - uses: msys2/setup-msys2@v2
+      with:
+        release: false
+        msystem: MINGW64
+        install: >
+          base-devel
+          mingw-w64-x86_64-toolchain
+        update: true
+
     - name: Build and test
       shell: bash
       run: |
         git clone https://github.com/steveicarus/ivtest.git
+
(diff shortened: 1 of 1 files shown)


steveicarus:
How does this PR relate to PR #398?

It appears that PR #398 is a draft, or holding pen. If that is the case, can you convert it into a draft, and can you close it when you are done using it as a reference?

And also, it this PR ready to merge?

umarcor:
> How does this PR relate to PR #398?

My end goal is to replace the current 'win' job (which is executed on MINGW64 only) with the two-job matrix in #398. There are currently several differences between them:

- Usage of action setup-msys2.
- Usage of a different entrypoint.
- Different procedure for calling build commands.
- Different sequence of test commands.

This PR is the first step towards closing the gap. At the same time, this PR is desirable anyway, as it provides caching for MSYS2 dependencies (the toolchain and base-devel).

> It appears that PR #398 is a draft, or holding pen. If that is the case, can you convert it into a draft, and can you close it when you are done using it as a reference?

I converted #398 to a draft.

> And also, it this PR ready to merge?

I think so!

Issue #399, opened 2020-11-29
CI


martinwhitaker:
It turns out that Travis CI is no longer automatically free to use for OSS projects - see

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
https://travis-ci.community/t/org-com-migration-unexpectedly-comes-with-a-plan-change-for-oss-what-exactly-is-the-new-deal/

We could try applying for OSS credits, but who knows how long that will last. One of our problems is that testing on Windows takes an inordinate amount of time, so uses a lot of credits.

An alternative is GitHub Actions, which is (currently) free for public repositories. I'll take a look at that.

umarcor:
To Do from #400:

> we need to compare against the appropriate expected results file (and generate version-specific expected results for MSYS2). Alternatively we could label all expected failures as such, and modify vvp_reg.pl to return a pass/fail exit code (as I did for vpi_reg.pl).

> Looks like we can create different scripts for different branches. We can use that to pass different expected results to the v11 branch.

steveicarus:
I agree with that approach. I was, in fact, thinking the same thing.

Make it so;-)

steveicarus:
Looks like we are switched over to Github Actions (PR #400). Can I close this issue?

umarcor:
@steveicarus, I'd rather keep it open until we implement the different tests depending on the branch. Anyway, feel free to close it.

umarcor:
It seems that #401 was executed in CI, merged, and executed in CI again, before latest changes were pushed to ivtest. As a result, although CI status is green, any CI run will fail due to missing files. See https://github.com/umarcor/iverilog/actions/runs/390884757.

I worked around it by fixing ivtest to an specific commit: https://github.com/umarcor/iverilog/runs/1472060125?check_suite_focus=true. Is that desired?

steveicarus:
I think, on the master branch, once a PR is merged, it doesn't matter. The test suite is going to advance as we add new fixes, so old commits will start to fail. I don't think it matters.

umarcor:
@steveicarus, please, restart https://github.com/steveicarus/iverilog/actions/runs/390778127 and you'll see what I mean. The first problem is `                   br_gh307: Failed - missing source file.`. That's because of the last commit in https://github.com/steveicarus/ivtest/commits/master.

steveicarus:
Fixed. You were correct that a file was missing.

martinwhitaker:
[...]

PR #404, opened 2020-12-01
ci/win: add 'build.sh'

umarcor:
In this PR, a custom entrypoint (`mingw64.cmd`) is used, instead of defining multiple commands through an envvar. This should be the last step before #398. Related to #399 and https://github.com/steveicarus/iverilog/pull/402#issuecomment-735533846:

- [x] Usage of action setup-msys2. Done in #402.
- [ ] Usage of a different entrypoint.
- [x] Different procedure for calling build commands. Done in this PR.
- [x] Different sequence of test commands. Done in #403.


steveicarus:
You've made a lot of PRs on this subject. Would it be better to mark this one as draft and only open it up to merging when the broader task is complete?

umarcor:
The point is that these updates are useful regardless of the migration to MSYS2 package builds (#398). As said above, this is the last one regarding the cleanup after migration from Travis to GitHub Actions. There were three PRs: one for the raw migration, another one for improving performance and this last one for improving readability.

Then, #398 might take longer because it seems that iverilog is not compatible with ASLR.

umarcor:
Closed in favour of #398.

PR #448, opened 2020-12-28
Add MacOS to CI/GHA

themperek:
It seems to fail during compilation at the moment,

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,34 @@ on:
 
 jobs:
 
+  mac:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [
+          '10.15'
+        ]
+    runs-on: macos-${{ matrix.os }}
+    name: '🍏 macos ${{ matrix.os }}'
+    steps:
+
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      run: |
+        brew install bison
+
+    - name: Build, install and check
+      run: |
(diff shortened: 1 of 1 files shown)


caryr:
This appears to be because of recent changes Steve made to support string parameters. It looks like he has added in C++11 constructs, but we do not check for or add the appropriate compiler flag to enable this. This is also failing on Ubuntu-16.04.

caryr:
I have pushed a fix that should resolve this. The issue was the usage of nullptr which is C++11. I have pushed a fix that should solve this:
```
-      if (pdims==nullptr) {
+      if (!pdims.get()) {
```
At least it fixed things on 16.04.

caryr:
pform_dump.cc did not error out so hopefully this will be good to go!

themperek:
Bison version is 3.7.4
```
bison --verbose -t --defines=parse.h -o parse.cc parse.y
parse.y:2.9-16: syntax error, unexpected identifier, expecting string
```

caryr:
It looks like this is the VHDL parse.y file and if I am reading the error line correctly it is the %define api.pure line and it doesn't like api.pure. No idea why that should not work with a modern bison. The same line is used in tgt-pcb/fp.y. Are you 100% sure this is a recent bison version? The only reference I found for this error message was when an outdated bison was used.

themperek:
Seems to work now.

umarcor (approved):

PR #469, opened 2021-01-04
Added parse-only support for pure virtual methods

svenka3:
Signed-off-by: Srinivasan Venkataramanan <svenka3@gmail.com>


caryr:
See my comments in #468. Do we also need a test that verify this is parsed correctly? What about failing test when they forget to define the method?

I'm making this a draft because I do not think it is ready to commit yet.

svenka3:
On "failing test" - a test can be written and added to regression, however the infrastructure needed to enforce this feature needs to be developed and am not sure if it can be done in parse.y alone - sure we can add a make_pure_virtual in property_qual.h

But to really ensure that pure virtual is implemented in a concrete/derived class, will require multiple file parsing and what I added for now is a start to that journey. 

Also I remember reading that Icarus today does not really check/do-the-right-thing for:
* virtual class (abstract class)
* virtual methods

Hence it will be harder to do these checks at this stage I guess. The best we can do is to add a clear warning.

Am not in a hurry to push this tiny check-in/PR, but rather saying we have to take smaller steps - IMHO.

Thanks

steveicarus:
Can you rebase this to the current master? The base of this branch does not include the updated regression test suite, and also missing a lot of work that has happened in the mean time. A rebase will help a lot.

parse.y
```diff
@@ -907,6 +914,11 @@ class_item /* IEEE1800-2005: A.1.8 */
   | method_qualifier_opt function_declaration
       { /* The function_declaration rule puts this into the class */ }
 
+  | pure_virtual_method_decl
+      { yywarn(@1, "OOP Pure Virtual methods are supported as parse-only. "
+                   "No enforcement is done for a concrete class to define these further");
+      }
+
```
steveicarus, line 921: We really need a "Sorry" message here, and not a warning. Since the feature is supported, it should generate an error, and a "Sorry" message (see examples elsewhere in the code) is the convention for expressing that.

Issue #436, opened 2020-12-20
SV: string len() operator in expressions not working


svenka3:
Hi,
  A string Q is used, its elements are processed inside a foreach, so far it works. Now, foreach element in this string Q, say we want to compare the length of the string and do some processing, Icarus breaks. 

Minmalist test:

```verilog
module m;

  string m_argv [$];

  initial begin
    m_argv.push_back ("str0");

    foreach (m_argv[i]) begin
      if(m_argv[i].len() >= 2) begin
        $display ("LARGE: %0d", lv_len);
      end
  
    end
  end
endmodule : m

```

```log
bb.sv:9: error: Scope index expression is not constant: i
bb.sv:9: error: Scope index expression is not constant: i
bb.sv:9: error: No function named `m_argv[i].len' found in this context (m.$ivl_foreach0).
bb.sv:9: error: Unable to elaborate condition expression.
4 error(s) during elaboration.
```

I tried breaking the code to get individual string, get it length, compare, that works as below.

```verilog
module m;

  string m_argv [$];
  string lv_s;
  int lv_len;

  initial begin
    m_argv.push_back ("str0");
    m_argv.push_back ("str1");
    m_argv.push_back ("str0123");

    foreach (m_argv[i]) begin
      $display ("%s", m_argv[i]);
      lv_s = m_argv[i];
      lv_len = lv_s.len();
      $display ("%0d", lv_len);
      if(lv_len >= 2) begin
        $display ("LARGE: %0d", lv_len);
      end
   
    end
  end
endmodule : m


```

caryr:
This all looks to be in code that Steve might be working on. I have the start of a patch, but for now I am releasing this. The fundamental issue is that the function call code does not know how to correctly elaborate indexed operations like identifiers can. I was able to get the test_width routines to work correctly since we just need to know the underlying array type (string in this case), but getting a NetESelect added to the function call elaboration is not obvious and I don't want to do a bunch of hacking if this is being reworked.

steveicarus:
I'll give this a try. It may be a big job, as it seems to be uncovering a few problems with name resolution in l-values.

caryr:
Do you want the work I already started or are you going to roll your own?

steveicarus:
What have you done? I'm first cleaning up some weird behaviors (i.e. strange type elaborations) Could you at least show me what you've done? Maybe I'll steal it.

caryr:
I'll send you the patch in email.

Issue #441, opened 2020-12-27
SV: assertion error: (darray), function elaborate_expr, file elab_expr.cc, line 2402.


svenka3:
I get below error:

```log
iverilog -g2012 ivl_uvm_i441.sv
Assertion failed: (darray), function elaborate_expr, file elab_expr.cc, line 2402.
sh: line 1: 34107 Done                    /usr/local/lib/ivl/ivlpp -L -F"/var/folders/rk/7dbwr18d6f9g7w9yqb06s6yh0000gn/T//ivrlg2222a612f" -f"/var/folders/rk/7dbwr18d6f9g7w9yqb06s6yh0000gn/T//ivrlg222a612f" -p"/var/folders/rk/7dbwr18d6f9g7w9yqb06s6yh0000gn/T//ivrli222a612f"
     34108 Abort trap: 6           | /usr/local/lib/ivl/ivl -C"/var/folders/rk/7dbwr18d6f9g7w9yqb06s6yh0000gn/T//ivrlh222a612f" -C"/usr/local/lib/ivl/vvp.conf" -- -
```

Code:

```verilog
package t;

  class c;
    virtual function c create (string name=""); 
    return null; 
  endfunction

  function c clone();
    c tmp;
    tmp = this.create();
   return(tmp);
endfunction


endclass : c
endpackage

module m;
  import t::*;

  initial begin
  end
endmodule 


```

Thanks

caryr:
This is the line that is failing `tmp = this.create();`

steveicarus:
OK, there are two bugs here. Icarus Verilog PECallFunction::elaborate_expr(...) is assuming the argument is a darray, which is not correct, and that is the assertion that we see. Relax that assertion and the deeper problem is that Icarus Verilog seems to be not handling the default argument. Call this.create(...) with an explicit argument, and it works. Rely on the implicit argument and it assertion in a different place.

Still working on it.

caryr:
I've tweaked the code a bit to resolve #440 and there could be conflicts so make sure you pull from head.

Issue #440, opened 2020-12-27
Cannot find class variables for null/class operator checks


svenka3:
Iverilog devel version, I get 

```log
XXXX Evaluate VEC4 expression (22)
```

Code below. Function create() has wrong return value leading to this.

```verilog
package t;

  class c;
    virtual function create (string name=""); 
    return null; 
  endfunction

endclass : c
endpackage

module m;
  import t::*;

  initial begin
  end
endmodule 

```

caryr:
This now correctly reports the incorrect usage for both devel and V11 for most cases not just the return you have here. The one remaining issue, and I will not close until that is fixed, is when you have mismatches for equality/non-equality check with class variables or null (e.g. null==0) is invalid, but we do not report a message for that yet.

@steveicarus @martinwhitaker I am having an issue when trying to get the information for class variables. It looks like the class scope does not have the variables so they are not found and test_width ends up just using some defaults that do not impact code generation, but is preventing me from adding the last check to verify that the equality/non-equality operators cannot check that both sides are class variables/null or are normal variables. There are comments in the test br_gh440.v that points to the actual source code where the check is commented out. Do either of you have an idea how to get this information?

steveicarus:
> This now correctly reports the incorrect usage for both devel and V11 for most cases not just the return you have here. The one remaining issue, and I will not close until that is fixed, is when you have mismatches for equality/non-equality check with class variables or null (e.g. null==0) is invalid, but we do not report a message for that yet.
> 
> @steveicarus @martinwhitaker I am having an issue when trying to get the information for class variables. It looks like the class scope does not have the variables so they are not found and test_width ends up just using some defaults that do not impact code generation, but is preventing me from adding the last check to verify that the equality/non-equality operators cannot check that both sides are class variables/null or are normal variables. There are comments in the test br_gh440.v that points to the actual source code where the check is commented out. Do either of you have an idea how to get this information?

Would the information you seek not be in the class definition (the class_def() method)?

caryr:
Yes, of course that was my thought, but it's not obvious how to get what is needed since this is not a scope so a bunch of extra work is needed. I'm also not certain the definition has the information in the correct format.

caryr:
I commented out the check in do_elab_and_eval(), netmisc.cc to resolve the issue in #445 until the class information is found correctly.

martinwhitaker:
[...]

Issue #438, opened 2020-12-27
SV: wait on boolean leads to Empty NexusSet from wait expression.


svenka3:
Using devel version I get:

```log
ivl_uvm_i439.sv:9: internal error: Empty NexusSet from wait expression.
ivl_uvm_i439.sv:9: Unable to elaborate statement in task $unit.ivl_uvm_mbx.get at ivl_uvm_i439.sv:8.
1 error(s) during elaboration.
```

Code below:

```verilog

`define IVL_UVM_MBX_T int

class ivl_uvm_mbx;
  bit is_empty;

  
  virtual task get (output `IVL_UVM_MBX_T out_xn);
    wait (is_empty == 0);
    /*
    while (is_empty == 1) begin
      #1;
    end
    */
  endtask : get 


endclass : ivl_uvm_mbx

module m;
  ivl_uvm_mbx mb_0;

  initial begin
    mb_0 = new();
  end
endmodule :m


```

PR #473, opened 2021-01-10
Queue element expressions

steveicarus:
This PR includes a rework of the symbol_search code, and the symbol_search() API. That function has accumulated hacks over the years, leading to a lot of ad hoc handling of things like members/methods of symbols. So I redesign the symbol_search() interface and largely reimplemented the function. It's now much cleaner and easier to read. There is also a compatibility API so that all that legacy use of symbol_search() can continue as before, but we should work towards using this new API.

This PR is still draft because I'm in the process of putting this new API to use to address issue #436. 

I'm publishing this as a draft PR so people can see what's going on here.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -111,6 +111,7 @@ O = main.o async.o design_dump.o discipline.o dup_expr.o elaborate.o \
     load_module.o netlist.o netmisc.o nettypes.o net_analog.o net_assign.o \
     net_design.o netclass.o netdarray.o \
     netenum.o netparray.o netqueue.o netscalar.o netstruct.o netvector.o \
+    make_ivl_type.o \
     net_event.o net_expr.o net_func.o \
     net_func_eval.o net_link.o net_modulo.o \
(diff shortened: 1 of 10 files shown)


caryr:
This is great news!

I believe, but have not confirmed the following are also failing for this same basic issue:

#441, #440, #438, #419, (maybe) #410 and #292

Please consider these when evaluating your changes.

steveicarus:
The PECallFunction::elaborate_expr method has been completely rewritten using the newer symbol_lookup. It's a bit cleaner looking, and more understandable. And in the process I've also fixed the issue that prompted this work. I expect I'll merge this work in the next day or so that it gets into master before someone else starts working in that general area. This might not be the last work, but it is better then what was there.

caryr:
Sounds good. I have started implementing let and doing a bunch of test cleanup after reviewing the results from sv-tests. I have tried to automate much of this based on defines so it should be transparent.

PR #505, opened 2021-05-04
Allow explicit logic to struct casting

phhorrein:
This was already supported with implicit cast, so it's really not a big change.

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -3442,6 +3442,9 @@ NetExpr* PECastType::elaborate_expr(Design*des, NetScope*scope,
 		  break;
 	    }
       }
+      if (dynamic_cast<const struct_type_t*>(target_)) {
+          tmp = cast_to_int4(sub, expr_width_);
+      }
       if (tmp) {
 	    if (tmp == sub) {
 		    // We already had the correct base type, so we just need to


martinwhitaker:
Do you have a regression test for this (and also for pull request #506)?

steveicarus:
Can you rebase this to the current master, which includes CI checks, and can you include a regression test program that tests this code? We won't merge this without a regression test, and rebasing to the current tip of master will get you a more convenient way to send regression tests.

steveicarus:
Since we never got feedback, in particular in the form of a sample that demonstrates the problem, I'm closing this PR. It is possible that this problem was fixed another way. In particular, this program, my best attempt at replicating the problem, seems to work fine:
```
module main;
   typedef struct packed {
      logic [3:0] hig;
      logic [3:0] low;
   } byte_t;

   typedef logic [7:0] uint8_t;
   
   byte_t   foo;
   uint8_t  bar;

   initial begin
      bar = 8'ha5;
      foo = byte_t'(bar);

      if (foo.hig !== 4'ha || foo.low !== 4'h5) begin
	 $display("FAILED - bar=%02h, foo={%1x, %1x}", bar, foo.hig, foo.low);
	 $finish;
      end

      foo.hig = 4'hb;
      foo.low = 4'h6;
      bar = uint8_t'(foo);
      
      if (bar !== 8'hb6) begin
	 $display("FAILED - bar=%02h, foo={%1x, %1x}", bar, foo.hig, foo.low);
	 $finish;
      end
      
      $display("PASSED");
      $finish;
   end
   
endmodule // main
```

PR #506, opened 2021-05-04
Adding struct packed array support for ports

phhorrein:
This is another simple pull request to add support for packed array with struct base type in module ports.

### DIFF
--- a/pform.cc
+++ b/pform.cc
@@ -2571,6 +2571,7 @@ void pform_module_define_port(const struct vlltype&li,
 			      bool keep_attr)
 {
       struct_type_t*struct_type = 0;
+      parray_type_t*packed_type = 0;
       enum_type_t*enum_type = 0;
       ivl_variable_type_t data_type = IVL_VT_NO_TYPE;
       bool signed_flag = false;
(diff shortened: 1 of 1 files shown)


steveicarus:
Can you rebase this to the current tip of "master", and add a test suite example to demonstrate and test it? We have reworked out continuous integration works in the iverilog repository, and rebasing will get this PR onto continuous integration. The PR otherwise looks fine, so it would be great to get it into the regular flow.

phhorrein:
Yes, I will do that as soon as I find some time, probably in a few days/weeks.

Le 2 février 2022 02:44:18 GMT+01:00, Stephen Williams ***@***.***> a écrit :

steveicarus (changes requested):
This PR no longer merges, so it needs to be rebased, and conflicts resolved.

Also, we need with this PR a regression test program that demonstrates the problem, and shows that this PR fixes it. That test program should be included in the regression test suite. If you need help with that, at least produce the program and we can give you guidance on how to include it in the CI test suite.

larsclausen:
I think this got resolved as part of bbe44deec2f4c0ba and can be closed.

PR #507, opened 2021-05-04
Support logical implication operator

zachjs:
Given that `constraint_expression`s are not supported, this change removes the conflicting production in favor of supporting the SystemVerilog logical implication operator. It would be possible to avoid the conflict by giving explicit precedence overrides, but I think it may make sense to avoid this until support for constraints is being added.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -1132,7 +1132,9 @@ constraint_declaration /* IEEE1800-2005: A.1.9 */
 constraint_expression /* IEEE1800-2005 A.1.9 */
   : expression ';'
   | expression K_dist '{' '}' ';'
-  | expression K_TRIGGER constraint_set
+  // FIXME: This causes a conflict with the logical implication operator
+  // production in `expression`.
+  // | expression K_TRIGGER constraint_set
   | K_if '(' expression ')' constraint_set %prec less_than_K_else
   | K_if '(' expression ')' constraint_set K_else constraint_set
   | K_foreach '(' IDENTIFIER '[' loop_variables ']' ')' constraint_set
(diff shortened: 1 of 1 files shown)


zachjs:
@martinwhitaker Does this seem reasonable?

martinwhitaker:
This will cause code with a complex constraint expression (i.e. one which is not an ordinary expression) to be rejected with the terse "syntax error" message, rather than the "sorry: Constraint declarations not supported." message. So I would prefer to fix the conflict. But I can't find a way to do it. The syntax looks to be inherently ambiguous.

zachjs:
Attempting to use the logical implication operator currently fails with the terse `syntax error` message. Valid use of the implication operator in a constraint expression currently fails with `sorry: Constraint declarations not supported`.

With this change as-is, the logical implication operator would be supported, but valid use of the implication operator in a constraint declaration would fail with `error: Errors in the constraint block item list`. However, the original more helpful error message would still be produced if the implication constraint expression can be parsed as an expression, which seems to be the common and natural usage.

This seems like a reasonable tradeoff to me. What do you think?

steveicarus:
@zachjs Can you rebase this to a later head, and include in your PR some test programs (in ivtest) to test your changes? The ivtest automated tests were merged into the iverilog repository for automatic CI in January, and this PR predates this. If you rebase your branch to the current head, then CI will better be able to regression test your PR, and it also becomes easier for you to add regression tests of your own for this PR.

steveicarus:
OK, I looked into this, and I found a way to resolve the shift/reduce conflict. (Ugh, it's ugly!) I created PR#614 which implements this parser fix, and also implements the runtime for this operator. It was never implemented. That's gotta say something about how often people use this operator.

The PR also creates a test program.

@zachjs If PR#614 is a suitable replacement, then I will close this PR and mark it obsolete.

steveicarus:
This PR is obsoleted by PR#614.

PR #511, opened 2021-05-12
Fix for cross compile build

mmicko:
Small explanation:
AX_PROG_CC_FOR_BUILD - moved from bottom to top of configure.ac, since all libraries tested for main compiler were also required in this case for native compiler, making a need of installation of zlib, bz2,... that were not used at all 

BUILDCC - previously wrong compiler was used for draw_tt, this change makes cross compile working fine.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -14,6 +14,7 @@ AC_CONFIG_HEADERS([tgt-pcb/pcb_config.h])
 AC_CANONICAL_HOST
 dnl Checks for programs.
 AC_PROG_CC
+AX_PROG_CC_FOR_BUILD
 AC_PREREQ([2.62])
 m4_version_prereq([2.70], [], [AC_PROG_CC_C99])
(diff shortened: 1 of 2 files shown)


sjkelly:
Would be nice to have this. We also hit this issue in: https://github.com/JuliaPackaging/Yggdrasil/pull/3446

sjkelly:
It would be good to backport to the v11_0 branch as well. We applied this on master and confirm this works well: https://github.com/JuliaBinaryWrappers/IVerilog_jll.jl/releases/tag/IVerilog-v12.0.0%2B0.

martinwhitaker:
Apologies for the delay in looking at this. It looks OK to me.

martinwhitaker:
Also backported to the v11 branch.

mmicko:
@martinwhitaker thanks a lot

PR #572, opened 2021-12-26
Fix unique0 keyword/token mapping

msinger:
I use `unique0` alot and always get the warning "value is unhandled for priority or unique case statement" when there is no default case (which there normally isn't in combination with `unique0`). I found that this is because the keyword `unique0` was mapped to the `K_unique` token instead of `K_unique0`.
I don't know if this was done intentionally to workaround another problem or if it was just a typo.


martinwhitaker:
Almost certainly a typo. The new keywords were added years before the parser was updated to handle them. Thanks for the patch.

martinwhitaker:
Applied in both the master and v11 branches.

PR #573, opened 2021-12-29
Short circuit logical operator

larsclausen:
Section 11.4.7 of the SystemVerilog standard states

```
The && and || operators shall use short circuit evaluation as follows:
  - The first operand expression shall always be evaluated.
  - For &&, if the first operand value is logically false then the second operand shall not be evaluated.
  - For ||, if the first operand value is logically true then the second operand shall not be evaluated.
```

vvp currently evaluates both operands of a logical operator. This works
fine as long as the right-hand side does not have a side effect. But if it has
the result might be incorrect.

E.g. for `a && b++` `b` must not be incremented if `a` evaluates to false.

The Verilog LRM mentions that it is allowed to short circuit any expression
"if the final result of an expression can be determined early". But there
is no requirement to do so.
    
So the new and the old behavior are both correct implementations in
Verilog.
    
Use the new behavior in both Verilog and SystemVerilog mode to make sure
the behavior is consistent when an expression has side effects.

Corresponding ivtest PR: https://github.com/steveicarus/ivtest/pull/17

### DIFF
--- a/eval_tree.cc
+++ b/eval_tree.cc
@@ -811,6 +811,36 @@ NetEConst* NetEBLogic::eval_arguments_(const NetExpr*l, const NetExpr*r) const
 
       const NetEConst*lc = dynamic_cast<const NetEConst*>(l);
       const NetEConst*rc = dynamic_cast<const NetEConst*>(r);
+
+      // If the left side is constant and the right side is short circuited
+      // replace the expression with a constant
+      if (rc == 0 && lc != 0) {
+	    verinum v = lc->value();
+	    verinum::V res = verinum::Vx;
+	    switch (op_) {
+		case 'a': // Logical AND (&&)
+		  if (v.is_zero())
+			res = verinum::V0;
+		  break;
+		case 'o': // Logical OR (||)
+		  if (! v.is_zero() && v.is_defined())
+			res = verinum::V1;
+		  break;
+		case 'q': // Logical implication (->)
+		  if (v.is_zero())
+			res = verinum::V1;
+		  break;
+		default:
+		  break;
+	    }
+	    if (res != verinum::Vx) {
+		  NetEConst*tmp = new NetEConst(verinum(res, 1));
+		  ivl_assert(*this, tmp);
+		  eval_debug(this, tmp, false);
+		  return tmp;
+	    }
+      }
+
       if (lc == 0 || rc == 0) return 0;
 
       verinum::V lv = verinum::V0;
--- a/tgt-vvp/eval_condit.c
+++ b/tgt-vvp/eval_condit.c
@@ -278,22 +278,39 @@ static int draw_condition_binary_le(ivl_expr_t expr)
 
 static int draw_condition_binary_lor(ivl_expr_t expr)
 {
+      unsigned label_out = local_count++;
+
       ivl_expr_t le = ivl_expr_oper1(expr);
       ivl_expr_t re = ivl_expr_oper2(expr);
 
       int lx = draw_eval_condition(le);
+      int tmp_flag = lx;
+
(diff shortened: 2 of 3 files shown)


martinwhitaker:
Actually the Verilog standard does mention short-circuiting - see 1364-2005 section 5.1.4.

https://www.accellera.org/images/eda/sv-bc/11114.html provides some more details - in particular note that in Verilog either operand can be evaluated first and short-circuit the other.

The behaviour required by SystemVerilog is compatible with traditional Verilog, but we should be wary about back-porting this change to v11, as it may break some users' existing code.

steveicarus:
Hmm, my first thought is that anyone writing Verilog code (not System
Verilog) who relies on short circuiting, or not short circuiting, is
asking for trouble, any anyone writing System Verilog are expecting
the patched behavior. With that in mind, maybe it makes sense to
accept this PR even for v11?

(I'm reading my e-mail in an airport, so I might be missing something...)

On Wed, Dec 29, 2021 at 3:39 PM martinwhitaker ***@***.***> wrote:
>
> Actually the Verilog standard does mention short-circuiting - see 1364-2005 section 5.1.4.
>
> https://www.accellera.org/images/eda/sv-bc/11114.html provides some more details - in particular note that in Verilog either operand can be evaluated first and short-circuit the other.
>
> The behaviour required by SystemVerilog is compatible with traditional Verilog, but we should be wary about back-porting this change to v11, as it may break some users' existing code.
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.
> You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>



-- 
Steve Williams
***@***.***

steveicarus:
OK, closer look at the patch and I certainly like the better factoring
in the code generator for logical operators. So the only question is
whether to accept the new behavior for the v11 branch. I say certainly
merge this for master, and probably try to cherry pick this into the
v11 branch.

On Wed, Dec 29, 2021 at 4:25 PM Stephen Williams ***@***.***> wrote:
>
> Hmm, my first thought is that anyone writing Verilog code (not System
> Verilog) who relies on short circuiting, or not short circuiting, is
> asking for trouble, any anyone writing System Verilog are expecting
> the patched behavior. With that in mind, maybe it makes sense to
> accept this PR even for v11?
>
> (I'm reading my e-mail in an airport, so I might be missing something...)
>
> On Wed, Dec 29, 2021 at 3:39 PM martinwhitaker ***@***.***> wrote:
> >
> > Actually the Verilog standard does mention short-circuiting - see 1364-2005 section 5.1.4.
> >
> > https://www.accellera.org/images/eda/sv-bc/11114.html provides some more details - in particular note that in Verilog either operand can be evaluated first and short-circuit the other.
> >
> > The behaviour required by SystemVerilog is compatible with traditional Verilog, but we should be wary about back-porting this change to v11, as it may break some users' existing code.
> >
> > —
> > Reply to this email directly, view it on GitHub, or unsubscribe.
> > Triage notifications on the go with GitHub Mobile for iOS or Android.
> > You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>
>
>
>
> --
> Steve Williams
> ***@***.***



-- 
Steve Williams
***@***.***

larsclausen:
> Actually the Verilog standard does mention short-circuiting - see 1364-2005 section 5.1.4.

Thanks, I've updated the description.

martinwhitaker:
Also applied to the v11 branch.

For the record, this doesn't cover logical operators in structural code. But given the way vvp works (pushing updated values through the structural netlist), I don't think we can fix that.

PR #574, opened 2022-01-03
Fix find_ivl_root for OSX

mmicko:
There is no /proc/self/exe on OSX, so it require a bit different code to get proper ivl_root.

### DIFF
--- a/driver/main.c
+++ b/driver/main.c
@@ -1044,7 +1044,7 @@ static void find_ivl_root(void)
 	   if the user has overridden $(bindir) or $(libdir), but there's
 	   not a lot we can do in that case.
  */
-#ifdef __MINGW32__
+#if defined(__MINGW32__)
       char tmppath[MAXSIZE];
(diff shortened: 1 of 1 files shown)


driver/main.c
```diff
@@ -1053,7 +1053,12 @@ static void find_ivl_root(void)
 	/* Convert to a short name to remove any embedded spaces. */
       len = GetShortPathName(tmppath, ivl_root, sizeof ivl_root);
 #else
+	#ifdef __APPLE__
```
martinwhitaker, line 1056: Would be better to use `#if defined(__MINGW32__) ... #elif defined(__APPLE__) ... #else ... #endif` instead of a nested `ifdef`.

mmicko: Agree, updated code.

martinwhitaker:
Thanks. Merged and backported to the v11 branch.

mmicko:
Thanks @martinwhitaker, those two were the only patches I had to do make it run on all platforms.

PR #581, opened 2022-01-15
Handle invalid struct members

larsclausen:
When something goes wrong when parsing a struct member, e.g. the type does
not exist, a nullptr is added to the struct member list. This will cause a
crash when iterating over the list.

E.g.

```
struct packed {
  logc x;
} s;
```

Add a check so that nullptr members are not added to the list.


steveicarus:
I have merged a change to the test flow, where the ivtest is bundled with the iverilog, in the ivtest subdirectory. It would be great if you can rebase this PR on the new master head, and fold your ivtest PR into this PR. That would make you a test case for this new arrangement. Please?

larsclausen:
> I have merged a change to the test flow, where the ivtest is bundled with the iverilog, in the ivtest subdirectory. It would be great if you can rebase this PR on the new master head, and fold your ivtest PR into this PR. That would make you a test case for this new arrangement. Please?

Done.

PR #590, opened 2022-01-17
Add support for explicit cast to packed array, packed struct and enum

larsclausen:
Currently explicit casts for packed types are only supported to atom2 and vector types. Generalize this code so that it supports cast to all packed types.

Correctly supporting explicit cast to enum is a bit more involved. Enum variables can only be assigned a value if the value is of the same enumeration type as the variable. Explicitly casting an expression to a enumeration type allows to assign it to an enum variable, even if the original type of the expression was not of that enumeration type.

In order to support this the type of the cast needs to be attached to the expression, so that the check in the assignment elaboration can verify the type.

### DIFF
--- a/PExpr.h
+++ b/PExpr.h
@@ -1007,6 +1007,7 @@ class PECastType  : public PExpr {
 
     private:
       data_type_t* target_;
+      ivl_type_t target_type_;
       PExpr* base_;
 };
 
--- a/dup_expr.cc
+++ b/dup_expr.cc
@@ -148,7 +148,7 @@ NetEConst* NetEConst::dup_expr() const
 
 NetEConstEnum* NetEConstEnum::dup_expr() const
 {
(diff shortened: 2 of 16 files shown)


steveicarus:
Are there any ILLEGAL casts that can be checked for in the test suite? I browsed the 1800-2017, and it seems pretty permissive. Might not be worth the effort to find examples, but if you have any handy, that would be great.

larsclausen:
packed types, strings and reals can be freely cast between each other, since you can also assign them to each other.

Unpacked arrays can be cast to each other if they have the same size and the base types are equivalent.

And then there is bitstream cast which allows unpacked structs, class, unpacked arrays and dynamic arrays to be cast to each other if they contain only packed types or strings and have the same size in number of bits.

That leaves things like anything with a real in it can only be cast to itself. Or a class with a class property.

Queues of strings to queues of reals should be forbidden if I understand it right.

larsclausen:
And maybe event. That can't be cast to nothing.

PR #591, opened 2022-01-18
[CI] Reduce build times for Mac and Ubuntu with ccache and GitHub cache.

MorriganR:
Reduce build times for Mac and Ubuntu with ccache and GitHub cache.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,6 +19,8 @@ jobs:
         ]
     runs-on: macos-${{ matrix.os }}
     name: '🍏 macos ${{ matrix.os }}'
+    env:
+      CCACHE_DIR: /Users/runner/ccache
     steps:
 
     - uses: actions/checkout@v2
@@ -27,13 +29,30 @@ jobs:
       run: |
         brew install bison
 
+    - name: restore ccache
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.CCACHE_DIR }}
+        key: ccache-mac-${{ matrix.os }}-${{ github.sha }}
+        restore-keys: |
+          ccache-mac-${{ matrix.os }}-
+
+    - name: ccache setup
+      run: |
+        mkdir -p ${{ env.CCACHE_DIR }}
+        brew install ccache
+        ccache -M 500M
+        ccache -v -s
+        ccache -z
+
     - name: Build, check and install
       run: |
-        export PATH="/usr/local/opt/bison/bin:$PATH"
+        export PATH="/usr/local/opt/ccache/libexec:/usr/local/opt/bison/bin:$PATH"
         autoconf
         ./configure
         make check
         sudo make install
+        ccache -v -s
 
     - name: Test
       run: ./.github/test.sh
@@ -49,6 +68,8 @@ jobs:
         ]
     runs-on: ubuntu-${{ matrix.os }}
     name: '🐧 Ubuntu ${{ matrix.os }}'
+    env:
+      CCACHE_DIR: /home/runner/ccache
     steps:
 
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


steveicarus:
Can you write up some details in the PR comment what, this is doing, what is being cached, and how it works? And maybe some run results to prove the point? This seems to be adding some complexity to the test workflow, so we would like a clearer picture of what, exactly this is doing. Thanks.

MorriganR:
Thanks for your question, it made me re-read the GitHub cache documentation.
I overlooked that GitHub cache uses not only the key but also the branch.
https://github.com/actions/cache#cache-scopes
https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key
This simplifies the test workflow.
I will try to add the necessary changes and clarifications as soon as possible.

steveicarus:
> Thanks for your question, it made me re-read the GitHub cache documentation. I overlooked that GitHub cache uses not only the key but also the branch. https://github.com/actions/cache#cache-scopes https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key This simplifies the test workflow. I will try to add the necessary changes and clarifications as soon as possible.

I've added a "Need Info" label to this PR. Let me know when you are ready for another look.
(You can also mark this PR as a draft until you are ready. Up to you.)

MorriganR:
This PR implements a fairly standard, for GitHub actions, scheme for accelerating the assembly of programs from source files. [Ccache](https://ccache.dev/) is used to cache compilation results while the job is running. [GitHub cache](https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows) is used to store the results and transfer them between workflow runs.

To enable ccache for Lin and Mac, just add the path to the directory in which ccache creates links to the beginning of the $PATH variable
```
export PATH="<path to ccache exec>:$PATH"
```

With Win(msys2) things don't look so transparent, because during the msys2 startup process, many [Environments](https://www.msys2.org/docs/environments/) were modified depending on the startup parameters.
The only workaround I've found is to change the CC and CPP values ​​just before running autoconf/configure/mingw32-make. The following code has been added to the PKGBUILD file:
```
  if [ "$GITHUB_ACTIONS" == "true" ]; then
    export CC="ccache gcc"
    export CXX="ccache g++"
  fi
```
this should have no effect on building the package on the local machine, but building on GitHub will use ccache.

steveicarus:
OK, this is interesting. I looked at the run times for the actions on this PR and compared it with the run times for other PRs, and the actions actually take considerably *longer* with this PR. Check out, for example, the run times for this PR: https://github.com/steveicarus/iverilog/pull/595

Do you have any thoughts on why this might be so?

MorriganR:
Using ccache can and should result in a slight [performance](https://ccache.dev/performance.html) hit on first run.

ThisPR

If you take a closer look at the test run log for this PR, you can see that GitHub cache failed to restore ccache, i.e. ccache had to re-create all cache files. This may explain the increase in the execution time of the `Build, check and install` step. But what I don't understand is the increase in time required to complete the `Test` step. It is possible that in both cases, other uncontrollable factors had an impact, for example, a drop in the performance of the VM instance on which the test was run.

In any case, GitHub cache has saved the ccache for this PR. If you now restart all tests for this PR (the `Re-run all jobs` button in the upper right corner), you will be able to see something like this:

reRUN

The proposed changes will not bring significant improvements for a project with a low PR intensity, less than one PR in seven days, in which case the GitHub cache for the master branch will be permanently outdated and deleted. Even so, the first test run for the new PR will create a ccache in the GitHub cache for the branch PR, which will be fairly useful if a change is made to the PR before it is merged into the main branch.

Also keep in mind that GitHub actions are free for all public repositories. Thus, the complication of the test launch script will not bring any financial benefit.

steveicarus:
I've rerun the actions, and I'm still not seeing a significant improvement, and possibly even a penalty. I would like @martinwhitaker  and @caryr to also take a look and see if they have an opinion on this.

PR #593, opened 2022-01-22
 Elaborate enums in the order they have been declared 

larsclausen:
enums for a scope are stored in a std::set. This means when iterating over
the enums during elaboration it is possible that they are elaborated in a
different order than they have been declared in. This causes problems if
one enum references items of the other enum. E.g.

```systemverilog
enum {
  A
} a;

enum {
  B = A
} b;
```

In the current implementation whether this works or not depends on the
pointer values of the enum_type_t for `a` and `b`, which can change between
environments.

To make sure that enums are always elaborated in the same order use a std::vector
instead of a std::set.

### DIFF
--- a/ivtest/regress-sv.list
+++ b/ivtest/regress-sv.list
@@ -237,6 +237,7 @@ enum_in_struct		normal,-g2005-sv	ivltests
 enum_in_class		normal,-g2005-sv	ivltests
 enum_in_class_name_coll	CE,-g2005-sv		ivltests
 enum_next		normal,-g2005-sv	ivltests
+enum_order		normal,-g2005-sv	ivltests
 enum_ports		normal,-g2005-sv	ivltests
 enum_test1		normal,-g2005-sv	ivltests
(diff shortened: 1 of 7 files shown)


ivtest/regress-sv.list
```diff
@@ -237,6 +237,7 @@ enum_in_struct		normal,-g2005-sv	ivltests
 enum_in_class		normal,-g2005-sv	ivltests
 enum_in_class_name_coll	CE,-g2005-sv		ivltests
 enum_next		normal,-g2005-sv	ivltests
+enum_order		CO,-g2005-sv		ivltests
```
steveicarus, line 240: Can you make this a "normal" program, even if you just add:
```
   initial $display("PASSED");
```
Yeah, that doesn't do much, but it does assure that the code generator doesn't choke on the change.

larsclausen: Added a PASSED that checks if all enum identifiers have the same value.

PR #594, opened 2022-01-22
 Allow unnamed begin/end block with only variable declarations

larsclausen:
While it is not a particular useful construct it is legal to have a
begin/end block with just variable declarations and no statements. E.g.

```systemverilog
begin
  int x;
end
```

At the moment there is a special rule for completely empty begin/end
blocks. Remove that rule and change the statement_or_null_list in the
begin/end block parser section to a statement_or_null_list_opt. This way it
covers both completely empty blocks as well as blocks with only variable
declarations.

Note that this already works as expected for named begin/end blocks

### DIFF
--- a/ivtest/regress-fsv.list
+++ b/ivtest/regress-fsv.list
@@ -82,6 +82,7 @@ pr3015421		CE			ivltests gold=pr3015421-fsv.gold
 resetall		normal,-Wtimescale	ivltests gold=resetall-fsv.gold
 scope2b			normal			ivltests
 sys_func_task_error	RE			ivltests gold=sys_func_task_error-fsv.gold
+unnamed_block_var_decl	normal			ivltests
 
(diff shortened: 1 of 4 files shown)


ivtest/regress-fsv.list
```diff
@@ -82,6 +82,7 @@ pr3015421		CE			ivltests gold=pr3015421-fsv.gold
 resetall		normal,-Wtimescale	ivltests gold=resetall-fsv.gold
 scope2b			normal			ivltests
 sys_func_task_error	RE			ivltests gold=sys_func_task_error-fsv.gold
+unnamed_block_var_decl	CO			ivltests
```
steveicarus, line 85: I would like to see this changed to "normal", and at least a stub $display("PASSED") in the test program.

larsclausen: Ok, added that.

PR #599, opened 2022-02-01
 Report error when trying to override non-existing parameter

larsclausen:
Overriding a parameter that does not exist will only generate a warning at
the moment. This can hide programming mistakes such as an typo in a
parameter override.

There is nothing in the LRMs to support that this should only be warning,
so elevate this to an error. This is consistent with how an error is
generated when trying to reference a non-existing port or variable.

The generated error message differentiates between whether the parameter
does not exist at all, or whether it is a localparam.

There are two regression tests that rely on that only a warning is
generated, these have been updated to expect an error.

### DIFF
--- a/elab_scope.cc
+++ b/elab_scope.cc
@@ -623,7 +623,7 @@ static void elaborate_scope_classes(Design*des, NetScope*scope,
       }
 }
 
-static void replace_scope_parameters(NetScope*scope, const LineInfo&loc,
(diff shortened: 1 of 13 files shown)


steveicarus:
Uh oh. Looks like when I merged this, the regression tests failed.
https://github.com/steveicarus/iverilog/actions/runs/1781685426

ivtest/ivltests/localparam_implicit.v fails in the merged head.

PR #615, opened 2022-02-14
Restrict edge event to SV

zachjs:
Is there interest in ensuring features like this are properly restricted to SystemVerilog? I've seen that there is some inconsistency here, so I'm curious what the desired behavior is.


steveicarus:
There generally is interest in keeping SV features out of non-SV compiles, but often that is handled by the keyword that triggers the feature only existing in SV, thus making the rule unreachable unless sv is enabled. That seems to not be the case in this example, and I'm surprised that Verilog(-2005) doesn't support that syntax.

Issue #665, opened 2022-04-01
Way to run vvp programatically, and receive call-backs to self?


hughperkins:
Hi,

I want something like the following architecture:

```
[c++ program]
       => contains vpi methods
       => launches vvp on a vvp file
       => vvp file calls back into the c++ program, via vpi methods, once a clock cycle
       => at some point the vvp run finishes, and control returns back to the c++ program
```

The idea is that my vvp file is a GPU under simulation, and my c++ program is like an OpenCL program, which contains an OpenCL kernel. The C++ program compiles the OpenCL kernel, and feeds that to the GPU, via vpi calls from the GPU simulator. Once the GPU has received the kernel, the GPU  can launch the kernel, finally returning the results to C++, via more vpi calls.

So, vpi works for me 🎉  The missing bits, I think (though I might find others), is:
- how to launch vpp programmatically, and
- how to have the vpp call back into the c++ program itself, i.e. treat the c++ program as a vpi module?

hughperkins:
(like, an ideal option might be if vvp can be compiled as a shared object, and linked to, like `-lvvp`, and the calling program can contain the methods that would normally go into the vpi file, and vvp would simply find those names within the exisiting process?)

hughperkins:
So, this is kind of a mess, but proof of concept at https://github.com/steveicarus/iverilog/compare/master...hughperkins:hp/poc-vvp-so?expand=1  It's a mess because this combines:
- modifications to vvp to make it build as libvvp.so (add lib_entry.cc ,and build libvvp.so)
- lib_entry.cc is pretty unclean
- the Makefile changes to build libvvp.so are pretty unclean
- it includes also demo/poc code to run this.

That said, it runs. It does assume that iverilog is built into `install.d`, at the root of the cloned iverilog repo. That being the case, to run the demo:
```
./configure --prefix=$PWD/install.d
make -j 8
make install
export PATH=$PWD/install.d/bin:$PATH
cd vvp
./run.sh
```

The output will hopefully resemble:
```
$ ./run.sh 
+ make
echo vvp exe ext vvp
vvp exe ext vvp
[...]

PR #667, opened 2022-04-02
Make possible to link with vvp, so can run simulation programmatically

hughperkins:
Make possible to link with vvp, so can run simulation programmatically, and have vpi call back into the same program.

See https://github.com/steveicarus/iverilog/issues/665


hughperkins:
Need to somehow add `-fPIC` on linux build. Not sure how to do this in an autoconf context?

hughperkins:
(added -fPIC hackily into Makefile.in for now...)

hughperkins:
(builds on both Mac OSX Big Sur, and Ubuntu 20.04, when I tried it)

steveicarus:
This PR has promise, but could use documentation. And this brings up the issue that there is need for documentation bundled with the source. Yours might be the thread that starts us down the path of creating a Documentation/ directory, much like how the Linux kernel source has a documentation directory.

To that end, can you add to your PR a file "Documentation/embedded-vvp.txt" that documents your feature, including how (and why) to use it?

hughperkins:
So, since creating this PR, I noticed that the license on your (very awesome :) ) project is GPL. This means that if I link my code with your project, and this PR is targeting just such a usage, taht would mean I would need to relicense my code to GPL, something I'm not very keen on doing.

Many projects recently use relatively permissive licenses, such as MIT. Pytorch has a permissive license, https://github.com/pytorch/pytorch/blob/master/LICENSE. Verilator is LGPL, which is right of the edge of being ok, but is probalby ok-ish, for my own usages, https://github.com/verilator/verilator/blob/master/LICENSE . Yosys has an unusual but permissive license, https://github.com/YosysHQ/yosys/blob/master/COPYING  My own projects are on various permissive licenses, such as MIT, https://github.com/hughperkins/VeriGPU/blob/main/LICENSE

Permissive licenses mean that for example Cadence could start to use iverilog under  the hood, if they chose. Giving you control over some aspects of Cadence software :) Depending on the exact license, they might or might not have to state clearly on their product that they have your product inside of it. Permissive licenses play nicely with code from other projects, and generally let people use the code for both non-commercial projects, and commercial projects.

steveicarus:
On the matter of Documentation: I've started this PR: https://github.com/steveicarus/iverilog/pull/675

On the matter of licensing, we have used explicit exceptions to allow for the sort of thing you are trying to do. Don't expect the license to change wholesale, since there are too many people involved who entered into this with certain expectations.

hughperkins:
> On the matter of licensing, we have used explicit exceptions to allow for the sort of thing you are trying to do. Don't expect the license to change wholesale, since there are too many people involved who entered into this with certain expectations.

Do you mean, you feel that the various contributors over the years might object to a wholesale change of license, but you feel might not object to making an explicit exception, whilst keeping the same overall license? How would an exception look like, in concrete actual writing?

joshtyler:
Is there any interest in continuing with this PR? Running a simulation programmatically would be a very useful feature to have.

hughperkins:
The license was not compatible with my own needs, but you're welcome to take ownership of this pr :)

gatk555:
I also have an application for this, and the documentation should be no problem.  I propose a short file, Documentation/usage/libvvp.rst, to say it exists, with most of the interface detail as comments in new header libvvp.h, so there is only one file to update.

In developing a prototype, I found some issues with the PR as it stands:
there is no way to pass text arguments to the simulation;
there is no way to test it without external code, not included with the PR;
and it creates a near-duplicate of main.cc that looks like a future maintenance headache.

That led me to re-work it, so there will be a new PR (but that would also be needed to add documentation).  In my version, main.cc is split, not copied.  Most of it becomes lib_main.cc and the new main.cc contains little more than main()
and the getopt() loop.  The changed code can be built two ways.

1. vvp is a small program that links the new libvvp.so.
2. vvp remains almost the same, but the new library is also built.

With the first choice, testing happens whenever vvp runs, so that looks attractive.  But now running from the build tree may require LD_LIBRARY_PATH, or use of chrpath or similar - not so attractive.  One possibilty: require "configure --with-libvvp" for option 1, otherwise vvp remains the same, library is not built.

I would appreciate guidance on documentation and build choices.

hughperkins:
@gatk555  Please go ahead and take ownership of the PR :) (and/or create a new one).

gatk555:
Thanks!

caryr:
@gatk555, with your libvvp work now merged into the repo is this pull request still relevant or has your pull subsumed this material?

gatk555:
@caryr, my apologies for the sluggish response: I have only just seen that.  Yes, I think this (#667) can now be closed.

G,

martinwhitaker:
Closed as superseded by #1068.

PR #669, opened 2022-04-03
 Handle signed return types for built-in methods of SV types

larsclausen:
The result for the built-in methods for the SystemVerilog types is
currently always unsigned. This can lead to incorrect behavior if the value
is sign extended or passed as an argument to a system function (e.g.
$display).

For most built-in methods this does not matter, since even though they have
a signed return type, they will not return a negative value. E.g. the
string `len()` or queue `size()` functions.

It does make a difference though for the queue `pop_front()` and
`pop_back()` methods. Their return type is the element type of the queue.
If the element type is signed and the value in queue is negative is will be
handled incorrectly.

E.g. the following will print `4294967295` rather than `-1`.
```SystemVerilog
int q[$];
q.push_back(-1);
$display(q.pop_front());
```

To correctly support this consistently assign the actual data type of the
built-in method's return value to the `NetESFunc`, rather than just the width
and base type. The width, base type and also the signedness can be derived
from the data type.

Note that assigning the actual data type is also required to support type
checking on the return value, e.g. as needed for enum types.

### DIFF
--- a/Statement.h
+++ b/Statement.h
@@ -249,9 +249,7 @@ class PCallTask  : public Statement {
 				      const char*sys_task_name) const;
       NetProc*elaborate_method_func_(NetScope*scope,
 				     NetNet*net,
-				     ivl_variable_type_t type,
-				     unsigned width,
-				     bool signed_flag,
+				     ivl_type_t type,
 				     perm_string method_name,
 				     const char*sys_task_name) const;
       bool test_task_calls_ok_(Design*des, NetScope*scope) const;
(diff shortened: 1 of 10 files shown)


martinwhitaker:
The title and description are wrong - this is about the standard methods for the various SV types which Icarus chooses to implement as (mostly private) system functions. The return types for ordinary system functions are handled correctly.

larsclausen:
Thanks. I've updated the title and description to `built-in methods of SV types`

martinwhitaker:
Thanks, that makes it clear.

larsclausen:
I'm going to close this. While it gets the initial signdness right it still has the wrong behavior when mixing signed and unsigned expression in arithmetic operations.

PR #694, opened 2022-04-18
Avoid signal look-up by name in the parser

larsclausen:
There are a few places in the parser where a signal is created and then
immediately afterwards the signal is looked up by name to make some changes to
the signal.

Refactor the code so that the pointer to the signal is retained and the look-up
by name is no longer necessary.


caryr:
@larsclausen I wanted to say I appreciate all the work you have put into cleaning up and refactoring the iverilog compiler lately.

caryr:
@larsclausen I wanted to say I appreciate all the work you have put into cleaning up and refactoring the iverilog compiler lately.

PR #720, opened 2022-05-22
 tgt-vvp: Fix out-of-bounds compressed assignment to arrays 

larsclausen:
If the index of an array access is known to be out-of-bounds during
elaboration it is replaced with 'x. In the tgt-vvp backend that is handling
compressed array assignments there is an assert() that triggers if the
index is an undefined immediate.

There is already an existing code path that is capable of handling
out-of-bounds access. Remove the assert and set the index to ULONG_MAX to
trigger taking the out-of-bound access path.

On this out-of-bounds path the write to the array is skipped. But this
leaves the result on the vector stack. Insert the `%pop/vec4` instruction
to make sure it is removed.

There is also an assert that gets triggered when the element value
of the array is wider than 32. Remove the assert and extend the
X value that gets pushed onto the stack to the right width using
the `%pad/s` instruction.

### DIFF
--- a/ivtest/ivltests/assign_op_oob.v
+++ b/ivtest/ivltests/assign_op_oob.v
@@ -1,0 +1,44 @@
+// Check that the assignment operator is supported for out-of-bounds indices.
+// The write should be skipped, but side effects of the right-hand side
+// expression should still get evaluated.
+
+module test;
+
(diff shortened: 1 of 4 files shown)


Igicjovana (commented):
[]()

PR #728, opened 2022-06-06
tgt-vvp: Fix syntax when using multiple events for non-blocking event control

larsclausen:
When multiple events are used in a non-blocking event control they need to
be combined into a single event using `event/or`.
    
The generated `event/or` statement is missing the trailing semicolon and
newline, which results in parser error when vvp tries to run.
    
E.g.

```SystemVerilog
event e, f;
integer x;
x <= @(e or f) 10;
```
    
Add the missing semicolon and newline to fix this.


steveicarus:
How was this bug not caught? I'm starting to think we need a code coverage target in the makefiles, since that snippet was clearly not executed. I bet there's lots of dead code in there, too.

PR #739, opened 2022-07-02
Add -R option to print runtime paths (for now, includedir only)

jevinskie:
This allows for, e.g., a Makefile to find the location of vpi_user.h in a portable manner. Similar to pkgconfig.


steveicarus:
Can you document this flag in the Documentation/usage/command_line_flags.rst file as well? This seems like a safe enough patch and I'm inclined to merge it. I appreciate that you updated the man page, but it would be best to get the online documentation covered as well.

jevinskie:
Done. Also matched the existing define naming convention and removed the quotes.

PR #740, opened 2022-07-04
cbNextSimTime: Don't call new cbNextSimTime events during current timestep

jevinskie:
Without a fix, registering a new cbNextSimTime callback within a cbNextSimTime callback will infinite loop and never advance the time.

It is my understanding that since cbNextSimTime is one-shot, I need to register a new cbNextSimTime callback within its own callback handler to get a callback for every sim tick. Is that correct?


steveicarus:
I think we need to see an example program that fails without this patch.

And as for regression tests, VPI regression tests are listed iin the ivtest/vpi_regress.list, and that includes some instructions on how to format VPI tests. If you are going to mess with the VPI implementation, let's be detailed.

steveicarus:
Also, note that this PR does not pass CI. You are breaking something. You'll need to figure that out before this can be merged.

jevinskie:
Thanks for the pointers! I do plan to add a minimal repro to the suite. It turns out the CI test was failing since the cbNextSimTime callback isn’t getting zero-initialized so the new check for a non-NULL time value in the CB struct passed and caused a segfault. I’ll fix that as well.

I’ll have to check to see if running the test with Valgrind would have caught it. Is a Valgrind run something we can add to the CI or is there a reason like CI timeouts that prevents it?

Is there any support in the test suite to compare results against a different simulator or a desire to add that (I’m willing to help)? I think that might be handy when testing part of VPI that aren’t so clearly specified.

caryr:
Valgrind is certainly too slow for CI. I used to run it every couple months to keep things clean, but I have not run it in over a year.

We test individual test using alternate simulators when possible, but we have never created infrastructure to test everything using an alternate simulator. In the past I have done this with a hacked test script, but some tests have Icarus specific constructs so fail. I don't remember the results for the VPI tests and I may have only run the VVP tests.

jevinskie:
I hope I've addressed your concerns. Without #741 first though, the ivlh_textio will now segfault.

caryr:
I'm not certain requiring the user to initialize the callback structure is correct since I do not believe that is required by the standard.

jevinskie:
Thanks, I’ll double check to see what Questa does since that’s the only other simulator I have available. I’ve had trouble reading the standard regarding ownership and allocation so I’m not surprised I messed that up.

jevinskie:
After reading this part of IEEE Std 1800-2012 (pg. 1065) I think it is required that the user allocates s_vpi_time. It doesn't even allow suppress/NULL.

> For reason **cbNextSimTime**, the time field in the time structure is ignored. The following fields shall need to be set before passing the `s_cb_data` structure to **vpi_register_cb()**:
>
> `cb_data_p->time->type`
>
> This field shall be set to either **vpiScaledRealTime** or **vpiSimTime**, depending on what time information the application requires during the callback. **vpiSuppressTime** (or `NULL` for the `cb_data_p->time` field) will result in an error.

caryr:
I'll need to look at the standard in more detail to say 100%, but your interpretation looks correct.

A couple minor comment. We do not normally indent both the case and the items in the case by the normal amount. The case is normally a smaller amount (say two more than the switch) and the enclosing items are a normal indent from the switch statement. Also do you need to add the error which should really be a sorry for vpiScaledRealTime since this is being ignored by cbNextSimTime?

jevinskie:
Thanks, I wasn't sure if the mix of tabs and spaces was intentional or accidental. I'll follow that style.

I'll just implement vpiScaledRealTime since now I know where it gets the scale from (the cb obj member).

I've got more concerns about ownership/allocation after reading:
> When a simulation time callback occurs, the application callback routine shall be passed a single argument,
which is a pointer to an `s_cb_data` structure [this is not a pointer to the same structure that was passed to
**vpi_register_cb()**]. The _time_ structure shall contain the current simulation time. The _user_data_ field shall be
equivalent to the _user_data_ field passed to **vpi_register_cb()**.

It looks like we're supposed to not reuse the s_cb_data passed to vpi_regiser_cb when calling its callback. I hope ownership isn't supposed to transfer to the callback so we can use a stack allocation instead of a heap allocation.

martinwhitaker:
This PR seems to have been overlooked, maybe because it still has the "Need info" label, which I'll remove. Sorry you haven't had a response before now.

Yes, the returned `s_vpi_time` structure needs to be allocated by the simulator. It might be easier to use `make_sync()` instead of `make_prepost()` for `cbNextSimTime` callbacks, as that does all the necessary allocation. The LRM states that ownership remains with the simulator.

Rather than your current solution, it would be simpler to do
````
    simulator_callback* next = NextSimTime;
    NextSimTime = 0;
````
then iterate through the `next` list.

The indentation of the switch statement still needs fixing.

I'll convert this to a draft PR, to show it still needs work.

PR #741, opened 2022-07-04
Fix sscanf overflows

jevinskie:
Without this fix, ivlh_textio was failing at step 11 where the time is read from the file. This was because clang happens to put period directly after the units buffer on the stack. sscanf writes the terminating NUL overwriting the low byte of period, which is 100, so it returns 0.

These were the only two points where clang identified an issue.

Example clang warning fixed:

warning: 'sscanf' may overflow; destination buffer in argument 4 has size 2, but the corresponding specifier may require size 3 [-Wfortify-source]

### DIFF
--- a/vpi/vhdl_textio.c
+++ b/vpi/vhdl_textio.c
@@ -188,7 +188,7 @@ static int read_vector(const char *string, s_vpi_value *val, vpiHandle var)
  */
 static int read_time(const char *string, s_vpi_value *val, PLI_INT32 scope_unit) {
     PLI_UINT64 period;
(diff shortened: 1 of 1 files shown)


steveicarus:
This looks good, what are the odds you can include a regression test that triggers the problem that this fixes?

jevinskie:
Step 11 of ivlh_textio is a regression test itself if built with -O0 -g and clang-14 on Ubuntu LTS 22.04. I’m not sure if different codegen is masking the issue on the Apple builds.

PR #745, opened 2022-07-26
Refactoring of main() and fixed potential bug regarding std::map[]

Mindar:
This is a small pull request to improve code quality: For now I only refactored the main() function to make it more readable. I plan to do further refactorings in this style in the future.

I also fixed a small (potential) bug with the usage of std::map's []-operator. See commit message for details.

My plan for the near-ish future is to do the following wherever it makes sense to do so:
- get rid of `goto` (found one in main, but there are dozens throughout the code)
- improve clarity of data-flow by reducing number of global variables
- get rid of unnecessary `extern` variables/functions/things through the use of headers
- split huge functions that do several different things into smaller functions that each do only one thing
- split huge files with several classes into several files with one class each (netlist.h contains declarations for 152 different classes)
- many other small things to reduce cognitive load during development

To accomplish this, I'd like to know which C++ version I should target? Is C++ 20 ok?

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -46,6 +46,7 @@ const char NOTICE[] =
 # include  <map>
 # include  <unistd.h>
 # include  <cstdlib>
+# include  <memory>
 #if defined(HAVE_TIMES)
 # include  <sys/times.h>
 #endif
@@ -60,6 +61,7 @@ const char NOTICE[] =
 # include  "compiler.h"
 # include  "discipline.h"
 # include  "t-dll.h"
+# include  "logging.h"
 
 using namespace std;
 
@@ -900,17 +902,15 @@ static void EOC_cleanup(void)
       filename_strings.cleanup();
 }
 
-int main(int argc, char*argv[])
-{
-      bool help_flag = false;
-      bool times_flag = false;
-      bool version_flag = false;
+static bool help_flag = false;
+static bool times_flag = false;
+static bool version_flag = false;
 
-      const char* net_path = 0;
-      const char* pf_path = 0;
-      int opt;
-
-      struct tms cycles[5];
+static const char* net_path = 0;
+static const char* pf_path = 0;
+static int opt;
+static struct tms cycles[5];
+int parse_cmdargs(int argc, char*argv[]) {
 
 #if defined(TRAP_SIGINT_FOR_DEBUG)
       signal(SIGINT, &signals_handler);
@@ -931,6 +931,9 @@ int main(int argc, char*argv[])
       flags["-o"] = strdup("a.out");
       min_typ_max_flag = TYP;
       min_typ_max_warn = 10;
+
+      // By default only errors are logged, i.e. LogSeverity::Error
+      uint8_t min_log_severity = 3;
 
       while ((opt = getopt(argc, argv, "C:F:f:hN:P:p:Vv")) != EOF) switch (opt) {
 
@@ -957,6 +960,10 @@ int main(int argc, char*argv[])
 	    break;
 	  case 'v':
 	    verbose_flag = true;
+          // A better implementation could increment a verbosity counter with every -v flag encountered
+          // the min_log_severity is then min(0, 3 - verbosity)
+          // That approach would allow to use more log levels for fine grained control of the output.
+          min_log_severity = 1;
 #          if defined(HAVE_TIMES)
 	    times_flag = true;
 #          endif
@@ -969,6 +976,8 @@ int main(int argc, char*argv[])
 	    flag_errors += 1;
 	    break;
       }
+
+      set_min_log_severity(min_log_severity);
 
       if (flag_errors)
 	    return flag_errors;
@@ -983,7 +992,7 @@ int main(int argc, char*argv[])
 
 	    dll_target_obj.test_version(flags["DLL"]);
 
-	    return 0;
+	    return -1;
       }
 
       if (help_flag) {
@@ -1005,7 +1014,7 @@ int main(int argc, char*argv[])
 "\t-V               Print version and copyright information, and exit.\n"
 
 		  ;
-	    return 0;
+	    return -1;
       }
 
       int arg = optind;
@@ -1015,8 +1024,8 @@ int main(int argc, char*argv[])
       }
 
       if (source_files.empty()) {
-	    cerr << "No input files." << endl;
-	    return 1;
+            log(LogSeverity::Error, "No input files.\n");
+            return 1;
       }
 
       separate_compilation = source_files.size() > 1;
@@ -1120,6 +1129,10 @@ int main(int argc, char*argv[])
       flag_tmp = flags["DISABLE_CONCATZ_GENERATION"];
       if (flag_tmp) disable_concatz_generation = strcmp(flag_tmp,"true")==0;
 
+      return 0;
+}
+
+int preprocess(){
 	/* Parse the input. Make the pform. */
       int rc = 0;
       for (unsigned idx = 0; idx < source_files.size(); idx += 1) {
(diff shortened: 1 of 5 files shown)


caryr:
One of the things we try to do with Icarus is make sure it compiles on many systems, even older ones. I would say we should never allow code that does not compile using the oldest Ubuntu LTS which is currently 18.04 (bionic beaver) which has gcc7.4/7.3 as the default depending on the target. The gcc 7.3 documentation reports it supports up to C++14.

Now when we consider Centos 7 which has gcc 4.8.5 and another 2 years of active support things get even worse where we are limited to C++11.

C++20 is certainly not going to be allowed since it will fail for many of the systems our users and even some of the developers commonly use. For maximum portability I would say we should stick to C++11 for a few more years to not adversely impact our users.

main.cc
```diff
@@ -971,7 +969,7 @@ int main(int argc, char*argv[])
       }
 
       if (flag_errors)
-	    return flag_errors;
+	    exit(flag_errors);
```
steveicarus, line 973: The refactoring is basically good, but I think better would be to have these new functions return int (not void) and leave the return statements instead of turning them into exit() statements.

main.cc
```diff
@@ -983,7 +981,7 @@ int main(int argc, char*argv[])
 
 	    dll_target_obj.test_version(flags["DLL"]);
 
-	    return 0;
+	    exit(0);
```
steveicarus, line 985: See above: rather return a value then exit().

main.cc
```diff
@@ -1379,6 +1382,19 @@ int main(int argc, char*argv[])
       return rtn;
 }
 
+int main(int argc, char*argv[])
+{
+	parse_cmdargs(argc, argv);
+
+	preprocess();
+
+	find_root_module();
```
steveicarus, line 1388: These functions should return error codes, and the main function should check those error codes.

caryr:
Also if you look at the machines we use for CI here on github you will see Ubuntu 18.04 so that will fail for C++20.

main.cc
```diff
@@ -1119,7 +1118,9 @@ int main(int argc, char*argv[])
 
       flag_tmp = flags["DISABLE_CONCATZ_GENERATION"];
       if (flag_tmp) disable_concatz_generation = strcmp(flag_tmp,"true")==0;
+}
 
+void preprocess(){
```
steveicarus, line 1123: Please return an error code in this and similar functions.

steveicarus (changes requested):
I agree that your changes make the code look better, but I have a few suggestions interspersed.

steveicarus:
It appears that this PR does not pass CI. You should check out what went wrong and fix it before we can merge it.

Thanks,

caryr:
One other thing, it looks like your spacing is a bit off. In the code we generally strive for six spaces, use a tab to represent 8 spaces where it makes sense (e.g. the first indentation is 6 spaces, the second is a tab plus four spaces, etc.) There are some exceptions for case labels which do not get their own indentation level. There is also some code that has smaller spaces because of large nesting.

larsclausen:
> One of the things we try to do with Icarus is make sure it compiles on many systems, even older ones. I would say we should never allow code that does not compile using the oldest Ubuntu LTS which is currently 18.04 (bionic beaver) which has gcc7.4/7.3 as the default depending on the target. The gcc 7.3 documentation reports it supports up to C++14.
> 
> Now when we consider Centos 7 which has gcc 4.8.5 and another 2 years of active support things get even worse where we are limited to C++11.

I'd love to be able to switch to C++11 unconditionally. Right now the CI still needs C++98 because that's the default version for some of the compilers if the version is not specfified. And we do things like have a #ifdef switch between unique_ptr and auto_ptr. If there is consensus that C++11 is OK as the minimum I'll try to prepare a PR that sets the compiler flags for it so CI will pass with C++11 code.

caryr:
@larsclausen is this a compiler switch issue or do you actually need to use/install an alternate version of the compiler to get C++11 support? If it's a switch issue then we should be able to add something to the configure script to get the correct flag added and fail if it is not available. Maybe I misread the documentation regarding what's supported by default versus gcc version. If this is the macos build then we should switch it to use 11 since 10.15 is now deprecated in the CI flow.

If we do make this change it should only be made in development and we should likely send out a message to see if anyone who is building from source would be negatively impacted by switching to C++11.

Mindar:
> It appears that this PR does not pass CI. You should check out what went wrong and fix it before we can merge it.

It has passed the tests on my machine. Not sure what's going on here. I'm running 5.18.10-1-MANJARO.
Is there any way to see the iverilog stdout for the specific testcases that failed?

steveicarus:
> The more I think about this, the worse it appears. Is it fine if I just indent using 6 spaces?

Use 6 spaces. That's fine.

Also, once we're all happy, if you can squash your commits that would be great too.

steveicarus (approved):
OK, LGTM.

main.cc
```diff
@@ -1212,171 +1218,194 @@ int main(int argc, char*argv[])
...
       while (!net_func_queue.empty()) {
-	    net_func func = net_func_queue.front();
-	    net_func_queue.pop();
-	    if (verbose_flag)
-		  cerr<<" -F "<<net_func_to_name(func)<< " ..." <<endl;
-	    func(des);
+            net_func func = net_func_queue.front();
+            net_func_queue.pop();
+            if (verbose_flag)
+            {
```
martinwhitaker, line 1316: This departs from the existing code style in two ways:
  - single statements aren't enclosed in braces
  - the opening brace is not put on a separate line

main.cc
```diff
@@ -1212,171 +1218,194 @@ int main(int argc, char*argv[])
...
-	    cout << "CALCULATING ISLANDS" << endl;
+            cout << "CALCULATING ISLANDS" << endl;
       }
       des->join_islands();
 
       if (net_path) {
-	    if (verbose_flag)
-		  cerr<<" dumping netlist to " <<net_path<< "..." <<endl;
+            if (verbose_flag)
+            {
```
martinwhitaker, line 1329: As above.

main.cc
```diff
@@ -1212,171 +1218,194 @@ int main(int argc, char*argv[])
...
 
- errors_summary:
-      if (! missing_modules.empty()) {
-	    cerr << "*** These modules were missing:" << endl;
+int main(int argc, char*argv[])
+{
+      // Note: parse_cmdargs() may return negative values if the operation was successful,
+      // but further execution of the program is not desired (e.g. after printing the help-text)
+      int errcode = parse_cmdargs(argc, argv);
+      if(errcode > 0) return errcode;
```
martinwhitaker, line 1399: We normally have a space after the `if` (repeated below)

martinwhitaker:
Regarding your questions:
1. I use MicroEMACS to edit code. I can set the tab key to perform 6-space indentation. When I finish making changes I have to use the entab-region command to manually convert to mixed spaces/hard tabs. Sometimes I forget! So yes, I agree the current style is awkward.
2. AFAIK, there isn't a coding style guide. When I started contributing, I just looked at the existing code and copied what I saw. One exception is the parse.y file, where there is an old style and a new style. I always convert to the new style when making changes there.

main.cc
```diff
@@ -900,18 +901,16 @@ static void EOC_cleanup(void)
...
-
-      struct tms cycles[5];
+bool help_flag = false;
+bool times_flag = false;
+bool version_flag = false;
 
+const char* net_path = 0;
+const char* pf_path = 0;
+int opt;
+struct tms cycles[5];
```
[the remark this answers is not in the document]
Mindar: Good catch, I simply copied them from main without thinking much.
My plan is to turn them into local variables again, once I've modified the newly created functions to take them as parameters.

caryr:
I guess over the years I have learned to mentally accommodate the spacing rules. I code it directly as seen and switch between different styles when working on different projects.

main.cc
Mindar: Will fix the first point once I'm back from vacation in 2 weeks. Regarding your second point, there are [counterexamples](https://github.com/steveicarus/iverilog/blob/master/main.cc#L1291) in the existing code. I'd strongly prefer to use that style and be more explicit about what's inside and outside the if-statement. It helps avoid bugs imo. Since the style throughout the file is already inconsistent, we might as well make use of the advantages of being a little more explicit.
Are you sure you want me to change that?

steveicarus:
This turned into a very different PR when you created the logging file. In Icarus Verilog, log/debug messages are enabled by enabling log message categories, not levels. I've found that levels just don't work. See the "debug_*" etc. in the compiler.h header file. So I'm not going to approve the min_log_severity additions. Sorry.

PR #767, opened 2022-09-14
Prevent invalid port redeclaration

larsclausen:
(System)Verilog allows to declare the port direction separate from the
signal declaration. E.g.

```SystemVerilog
output x;
integer x;
```

But this is only allowed if the port declaration
 * does not have an explicit net type
 * does not have an explicit data type
 * is a non-ANSI style declaration

For all other cases of port declarations the signal is considered fully
defined and it is not allowed to have a separate signal declaration.

In addition the declared packed dimensions need to match between the port
and signal declaration.

In the current implementation there are a few cases where this is not
handled correctly.

1) It is possible to declare non-ANSI task ports with the same name over
and over again, if it was declared as a signal before the port.

```SystemVerilog
task t;
  string x;
  input logic x;
  output real x;
endtask
```

2) It is possible to re-declare non-ANSI input ports of a module that have
a data type, but no explicit net type.

```SystemVerilog
module M;
  input integer x;
  wire integer x;
endmodule
```

3) It is possible to re-declare a ANSI port if it has an implicit data type.

```SystemVerilog
module M(output [1:0] x);
  reg [1:0] x;
endmodule
```

4) It is possible to declare a vector signal for a scalar non-ANSI task
port.

```SystemVerilog
task t;
input x;
reg [7:0] x;
```

To handle all of these correctly refactor signal declaration and lookup a
bit.

The PWire class that represents a signal already has two flags `port_set_`
and `net_set_`. These flags indicate whether a signal has already been used
in a port or signal declaration. A port declaration that includes an
explicit data type is considered both a port and signal declaration.

Use these flags to decide whether it is possible to extend an existing
declaration. E.g. when creating a port without an explicit data type and a
PWire by that name already exists and the `port_set_` flag is not set
extend the existing PWire. On the other hand if the `port_set_` flag is
already set report an error.

Similar for signals but with the `net_set_` flag.

For port declarations with an explicit data type or ANSI style port
declarations it is always an error if a PWire by that name already exists.

This is for both module and task/function ports.

There is a small number of existing regression tests that is not compliant
with the new strict behavior.

There is one test which re-declares ANSI ports as variables. This test is
not standard compliant and was added in [response to a similarly
non-compliant test in the Yosys test suite](https://sourceforge.net/p/iverilog/mailman/message/34182256/). That test has sub-sequentially
been [updated to be standard compliant.](https://github.com/YosysHQ/yosys/issues/1570)

There are a few tests that mix scalar ports with non-scalar signals. These
tests are updated to run with the existing no-io-range-error flag, which turns
the error into a warning.


martinwhitaker:
> There is one test which re-declares ANSI ports as variables. That test is
updated to use non-ANSI ports.

Better to just delete it, as it is only there to test the behaviour you are disallowing. See https://sourceforge.net/p/iverilog/mailman/message/34182256/

larsclausen:
> > There is one test which re-declares ANSI ports as variables. That test is
> > updated to use non-ANSI ports.
> 
> Better to just delete it, as it is only there to test the behaviour you are disallowing. See https://sourceforge.net/p/iverilog/mailman/message/34182256/

Interesting. This is form the SV LRM 23.2.2.2 ANSI style list of port declarations

> Each port declaration provides the complete information about the port. The port’s direction, width, net or
variable type, and signedness are completely described. **The port identifier shall not be redeclared, in part or
in full, inside the module body.**

I think that's pretty clear that it is not allowed. The Verilog LRM has the same paragraph, but only up to the `completely described` part, it is lacking the clarification the re-declaration is not allowed. We could allow this in Verilog mode for backwards compatibility since there appears to be code out there that uses it. Or we could add a flag similar to `no-io-range-error` to turn this into a warning.

martinwhitaker:
Now I read 1364-2005 again, I think I was wrong to accept that behaviour. And I see the yosys test was changed accordingly in 2019.

larsclausen:
> Now I read 1364-2005 again, I think I was wrong to accept that behaviour. And I see the yosys test was changed accordingly in 2019.

Ok, I've removed the test with a comment why it was originally added and why it is not standard compliant.

PR #780, opened 2022-10-06
Prevent non-blocking writes to fields of automatic structs

larsclausen:
The `PEIdent::has_aa_term()` method still uses the old `symbol_search()`
and will fail to find the variable if part of the identifier path is a
member select of a variable.

As a result such writes to fields of automatic structs can be classified as
static and it is possible to do non-blocking assignments to them. E.g.

```SystemVerilog
task automatic t;
  struct packed {
    logic x;
  } s;
  s <= ...; // This fails
  s.x <= ...; // This works, but should fail
endtask
```

Switch to the new symbol search to make sure this case is handled
correctly. The new symbol search will correctly handle identifier paths
that have a trailing item after the variable, while the old symbol search
will always return an error in that case.

Note that while it is not allowed to do a non-blocking write to a class
object automatic variable, it is allowed to do a non-blocking write to a
property of a class object that is stored in an automatic variable, as the
non-blocking write is supposed to capture a reference to the object and not
reference the variable. E.g.

```SystemVerilog
class C;
  int x;
endclass

task automatic t;
  C c;
  c <= ...; // Not allowed
  c.x <= ...; // Allowed
endtask
```

Non-blocking access to class properties is not yet support in
Icarus in general, but the error handling for that needs to be done
somewhere else.

### DIFF
--- a/PExpr.cc
+++ b/PExpr.cc
@@ -446,19 +446,17 @@ void PEIdent::declare_implicit_nets(LexicalScope*scope, NetNet::Type type)
 
 bool PEIdent::has_aa_term(Design*des, NetScope*scope) const
 {
-      NetNet*       net = 0;
-      ivl_type_t    cls_val;
-      const NetExpr*par = 0;
-      ivl_type_t    par_type;
-      NetEvent*     eve = 0;
+      symbol_search_results sr;
+      if (!symbol_search(this, des, scope, path_, &sr))
+	    return false;
 
-      scope = symbol_search(this, des, scope, path_, net, par, eve,
-                            par_type, cls_val);
+      // Class properties are not considered automatic since a non-blocking
+      // assignment to an object stored in an automatic variable is supposed to
+      // capture a reference to the object, not the variable.
+      if (!sr.path_tail.empty() && sr.net && sr.net->class_type())
+	    return false;
 
-      if (scope)
-            return scope->is_auto();
-      else
-            return false;
+      return sr.scope->is_auto();
 }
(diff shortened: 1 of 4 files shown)


caryr:
Is this allowed? It seems like a non-blocking write could perform the actual write to the structure/variable after the task has completed and hence the memory it is writing to could/would have already been deallocated.

larsclausen:
> Is this allowed? It seems like a non-blocking write could perform the actual write to the structure/variable after the task has completed and hence the memory it is writing to could/would have already been deallocated.

Which part? NB assignment to a struct member of an automatic variable is not allowed, but it is not detected as an error in the current code base and the runtime will just crash. This patch makes sure an error is reported for this case.

NB assignment to a class property is always allowed since 1800-2012, before it was always forbidden. The NB assignment is supposed to hold a reference to the object and only release it once the NB assignment has completed. My assumption is that this was added so you can do NB assignments from class methods where you do not know if `this` is an automatic or static variable.

Icarus does not support NB assignments to class properties yet. But that should be a `sorry` message, not a `error` message.

caryr:
I misread your original comment/code. I thought your patch would allow a non-blocking assignment to automatic variables. Based on rereading, your latest comments and believing the title :) I see you just found a hole in the existing check. Note this really should check the variable and not the scope, but I can't remember if that infrastructure is in place yet, since an automatic variable can be created in a static scope using the automatic keyword and a static variable can be created in an automatic scope.

Sorry for the noise.

Issue #793, opened 2022-11-28
Compilation terminated by assertion error


cemery123:
Hi When I used Iverilog, an assertion error occurred and the compilation was terminated. I can't figure out where the specific problem is, the error message is as follows
`ivl: net_link.cc:130: Nexus* Link::find_nexus_() const: Assertion 'next_' failed.`
Wondering if you could give me some info on this.

larsclausen:
Hi, thanks for reporting the issue. Can you provide an small example code snippet that is able to reproduce to the issue. This will help to better understand what is causing the issue.

cemery123:
Hi My code snippet may be a bit long, I'm uploading it now and I hope it won't have too much extra impact on you

cemery123:
This is my design file：
```
module top  (y, clk, wire3, wire2, wire1, wire0);
  output wire [(32'h89):(32'h0)] y;
  input wire [(1'h0):(1'h0)] clk;
  input wire signed [(4'hb):(1'h0)] wire3;
  input wire [(4'hb):(1'h0)] wire2;
  input wire signed [(4'h8):(1'h0)] wire1;
  input wire signed [(4'h9):(1'h0)] wire0;
  wire signed [(4'hb):(1'h0)] wire22;
  wire signed [(3'h7):(1'h0)] wire21;
  reg signed [(2'h3):(1'h0)] reg20 = (1'h0);
  reg [(4'h9):(1'h0)] reg19 = (1'h0);
  reg [(4'ha):(1'h0)] reg18 = (1'h0);
  reg [(4'h8):(1'h0)] reg17 = (1'h0);
  reg signed [(3'h7):(1'h0)] reg16 = (1'h0);
  reg [(2'h3):(1'h0)] reg15 = (1'h0);
  reg [(4'h8):(1'h0)] reg14 = (1'h0);
  wire signed [(3'h5):(1'h0)] wire13;
  reg [(4'h8):(1'h0)] reg12 = (1'h0);
  reg [(3'h6):(1'h0)] reg11 = (1'h0);
  reg [(4'h9):(1'h0)] reg10 = (1'h0);
  reg [(3'h7):(1'h0)] reg9 = (1'h0);
  reg signed [(3'h5):(1'h0)] reg8 = (1'h0);
  wire [(3'h5):(1'h0)] wire7;
  wire [(3'h7):(1'h0)] wire6;
  wire signed [(4'h9):(1'h0)] wire5;
[...]

PR #795, opened 2022-11-30
 Insert drivers for undriven nets

larsclausen:
Trying to add a drive strength or delay to a undriven net will result in an
assertion. Make sure that a driver is added to undriven nets.

A driver is already added for all NetESignals, which covers most expression that
can produce a raw net rvalue. But there are other ways we can end up with just a
net as the rvalue, e.g.  when applying a sign cast to a net. The following
example triggers the issue

```SystemVerilog
wire [7:0] a;
wire [7:0] b = $signed(a);
```

This addresses #793


larsclausen:
I'm a bit uncertain about whether this is the right approach. We could also just ignore delays and drive strengths on assignments from undriven nets, since they shouldn't have any effect anyway.

PR #799, opened 2022-12-06
 Allow to omit dimensions in foreach loop

larsclausen:
SystemVerilog allows to skip dimensions in a foreach loop by not specifying
an identifier name for the dimensions. E.g. the following will iterate over
the first and last dimensions, but skip the middle dimension.

```SystemVerilog
int x[2][3][4];
foreach(x[a,,b]) $display(a, b);
```

The output will be
```
          0          0
          0          1
          0          2
          0          3
          1          0
          1          1
          1          2
          1          3
```

Add support for this to the parser as well as elaboration.

Also report an error when more loop variables are specified than
the array has dimensions. Currently this results in an assert.


ivtest/regress-sv.list
```diff
@@ -600,6 +600,10 @@ sv_foreach2		normal,-g2009		ivltests
 sv_foreach3		normal,-g2009		ivltests
 sv_foreach4		normal,-g2009		ivltests
 sv_foreach5		normal,-g2009		ivltests
+sv_foreach6		normal,-g2009		ivltests
+sv_foreach7		normal,-g2009		ivltests
+sv_foreach8		normal,-g2009		ivltests
```
steveicarus, line 605: I think it would be handy to include a gold file for the sv_foreach8 test that shows not only that the right number of iterations are taken, but that the indices are correct. The sample output in the PR comment would be a good example of what I'm thinking.

larsclausen: Done

PR #832, opened 2022-12-26
 ci: Windows: Add gperf to dependency list

larsclausen:
msys2 [removed the gperf package from the base-devel](https://github.com/msys2/MSYS2-packages/commit/085698dce3b5e481f0db586173d0157c8e5c8ff9) package collection.

Add gperf explicitly to the dependency list, without this the Windows
CI will fail.

Also fix the following warning `=> WARNING: MINGW_INSTALLS is deprecated, use MINGW_ARCH instead`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,6 +98,7 @@ jobs:
         install: >
           git
           base-devel
+          gperf
           mingw-w64-${{ matrix.arch }}-toolchain
 
     - name: Build and check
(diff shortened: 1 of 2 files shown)


steveicarus (approved):
SoaB! I just saw ctions starting to fail. Thanks for getting on top of this.

larsclausen:
All of the builds are now failing with ```Error: VVP input file 12.0 can not be run with run time version 13.0 (devel)``` I think we need to bump the version in all the vvp/examples files to make the CI work again.

PR #839, opened 2022-12-28
Elaborate member selects in the scope where they are used

larsclausen:
There are a few cases where a member select on a package scoped identifier
is evaluated in the scope of the package rather than the scope where the 
identifier is referenced.

This leads to incorrect behavior if a local symbol is used as an index in a
part select of the referenced member select. E.g.

```SystemVerilog
package P;
  localparam N = 1;
  struct packed {
    logic [3:0] x;
  } s = 4'b0101;
endpackage

module test;
  localparam N = 2;
  initial $display(P::s.x[N]); // Will print 0, should print 1
endmodule
```

Use the scope where the member select is used, rather than the scope where
the identifier is defined, to fix this.

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -4369,7 +4369,7 @@ NetExpr* PEIdent::elaborate_expr(Design*des, NetScope*scope,
 
       if (!sr.path_tail.empty()) {
 	    if (net->struct_type()) {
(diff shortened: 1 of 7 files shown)


steveicarus:
Looks like there are some conflicts between this PR and a previously merged PR of yours. Looks like a pretty simple fix, in ivtest/regress-sv.list.

Issue #827, opened 2022-12-22
Ternary operator with constants to $display()


bpringlemeir:
Below is a test case.

```
module test();
   reg expected = 1;
   initial begin
      $display("Const Positive %s", 1 ? "yes" : "NO");
      $display("Const Negative %s", !1 ? "yes" : "NO");
      $display("Var Positive %s", expected ? "yes" : "NO");
      $display("Var Negative %s", !expected ? "yes" : "NO");
      $finish;
   end
endmodule

```

I am using the WSL2 with Ubuntu 20.04 release of the iverilog and vvp.

```
$ iverilog -V | head -n 1
Icarus Verilog version 10.3 (stable) ()
Unable to get version from "/usr/lib/x86_64-linux-gnu/ivl/ivlpp -V"
Unable to get version from "/usr/lib/x86_64-linux-gnu/ivl/ivl -V -C"/tmp/ivrlh337f554a" \
    -C"/usr/lib/x86_64-linux-gnu/ivl/vvp.conf""
$ iverilog -Wall -g2005-sv test.v
$ vvp a.out
Const Positive yes
Const Negative
Var Positive yes
Var Negative  NO
```

The **Const Negative** line does not supply the string 'NO'.  If the constants are verilog defines, then the use case becomes more apparent.

martinwhitaker:
Note that version 10.3 is very old and no longer supported. But this bug is still present in the master branch.

bpringlemeir:
I don't need to have it for 10.3, I have a work around.  I was just reporting in case it was still present.

bpringlemeir:
I reproduced with 22.04 version 11.0 release on stock Linux.
I recompile with 12 development master and reproduced.

The VVP files all have,

` %vpi_call/w 2 5 "$display", "Const Negative %s", "\000NO" {0 0 0};`

Is it something to do with the parsing of "NO" or why could a NUL character be added by some sort of optimization pass?

larsclausen:
The rules for ternary operators are that both arguments are expanded to the width of the widest argument. In the non-optimized case a temporary variable that is a vector gets generated. E.g.
```SystemVerilog
reg [23:0] v;
if (expected) v = "Yes"; else v = "No";
$display(..., v);
```
When assigning a string literal to a vector that is wider it gets left padded with 0s. $display then ignores those extra 0. This is the correct behavior.

[...]

PR #851, opened 2023-01-02
 vvp: Remove null-bytes when converting string literals to SV strings

larsclausen:
The `%pushi/str` and `%concati/str` instructions should remove null-bytes
from the string literal when converting it to a SV string. This is defined in
section 6.16 ("String data type") of the LRM (1800-2017).

This is already handled correctly when converting a vector from the stack
to a SV string, just not when converting a string literal to SV string.

### DIFF
--- a/ivtest/ivltests/sv_string8.v
+++ b/ivtest/ivltests/sv_string8.v
@@ -1,0 +1,43 @@
+// Check that null-bytes are removed when converting to string type.
+
+module test;
+
+  reg failed = 1'b0;
+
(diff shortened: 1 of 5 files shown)


caryr:
Does the VPI interface also work correctly when given a string with NULL bytes?

larsclausen:
SystemVerilog string type strings are not allowed to contain a null-byte, so for them there is nothing special to do with the VPI API.

For passing string literals to VPI I'll open a separate PR that fixes the issues of #827.

Issue #482, opened 2021-01-18
assigning a 2D array of integer pairs


logicmonkey:
I would like to be able to initialise a 2D array of integer pairs with arr[p][q] = {int_a, int_b}.
Is this possible using -g2005-sv ?
Thanks!
`
module intarr();
  integer arr[7:0][3:0][1:0];
  initial begin
    for (int ii=0; ii<8; ii++)
      for (int jj=0; jj<4; jj++) begin
        arr[ii][jj] = {jj, ii};                // error
        // arr[ii][jj][0] = ii; // This...
        // arr[ii][jj][1] = jj; // ...works
      end
  end
endmodule
`

logicmonkey:
I get a failure with -g2012 too (which I guess is what I should be selecting anyway if I want SystemVerilog features?)

So aside from the feature above, is there a formal stance on which standard will be best supported going forward? It was mentioned to me somewhere that Icarus SV development had halted due to the size of the task.

caryr:
The generation should mostly just control the keywords that are used during parsing and trigger support for SystemVerilog in general.

You have been given bad information regarding our development. SV development is slow in Icarus because we have a small group of developers, but we regularly fix issues and add new features. It was going at a faster pace previously when some additions were being sponsored and we were not as busy with our paying jobs.

This is likely a bug/enhancement in Icarus, but I have not actually checked the code.

logicmonkey:
Very glad to hear it - this was pretty much my response, but it's nice to have that confirmed.
Apologies for piggy-backing that onto a ticket.

caryr:
I can confirm this is an issue. I believe the issue is we do not currently support assigning unpacked arrays. I'm going to mark this as an enhancement since you have a viable work around.

PR #868, opened 2023-02-04
Add initial support for packed array and packed struct assignment patterns

larsclausen:
SystemVerilog allows to use assignment patterns to assign a value to a
packed array or packed struct typed variable.

This is similar to using a concatenation, with the difference that for
concatenations the values are evaluated in a self-determined context and
for assignment patterns they are evaluated in a context defined by the
element type of the packed array or the member type of the struct.
This means that the value is for example automatically width expanded or
truncated if it does not have the same size as the element type. Automatic
type conversion is also done when allowed. E.g.

```SystemVerilog
bit [3:0][3:0] x = '{1'b1, 32'h2, 3.0, "TEST"};
$display("%x", x); // -> 1234
```

Nested assignment patterns are also supported. E.g.

```SystemVerilog
struct packed {
  byte x;
  bit [2:0][3:0] y;
} x = '{1, '{2, 3, 4}};
$display("%x", x); // -> 1234
```

Add support for using assignment patterns as the right hand side value.
Since the complete type of the target variable is required to correctly
evaluate the assignment pattern it is handled as a special case in
`elab_rval_expression()`. For other types of expressions for packed values
only the total width of the target value is provided to the rvalue
elaboration function.

SystemVerilog also supports assignment patterns for the left hand side in
assignments. This is not yet supported.

Also not yet supported is specifying array elements by index, struct members by
name as well as the `default` keyword. Such as
```SystemVerilog
bit [3:0][3:0] x = '{1:1, default: 0};
```

### DIFF
--- a/PExpr.h
+++ b/PExpr.h
@@ -213,8 +213,15 @@ class PEAssignPattern : public PExpr {
 				     unsigned expr_wid,
                                      unsigned flags) const;
     private:
-      NetExpr* elaborate_expr_darray_(Design*des, NetScope*scope,
-				      ivl_type_t type, unsigned flags) const;
+      NetExpr* elaborate_expr_packed_(Design *des, NetScope *scope,
+				      ivl_variable_type_t base_type,
+				      unsigned int width,
+				      const std::vector<netrange_t> &dims,
+				      unsigned int cur_dim,
+				      bool need_const) const;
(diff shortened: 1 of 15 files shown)


logicmonkey:
Great fixes. Would this fix #482 ?

larsclausen:
> Great fixes. Would this fix #482 ?

Not yet, but that's next.

PR #875, opened 2023-02-27
ivtests: Add support for vlog95 tests

steveicarus:
This adds the "normal-vlog95" test type, because we need to run a different set of commands for this to work so a new test type makes sense. This also moves a few of the existing tests to the new format in order to test the new test rig.

This addresses issue#874


ivtest/run_ivl.py
```diff
@@ -240,8 +199,99 @@ def do_run_normal(options : dict, expected_fail : bool) -> list:
         return [1, "Failed - No PASSED output, and no gold file"]
 
 
+def do_run_normal_vlog95(options : dict, expected_fail : bool) -> list:
+    '''Run the iverilog and vvp commands.
+
+    In this casem, run the compiler with the -tvlog95 flag to generate
+    an intermediate verilog file, then run the compiler again to generate
+    a vvp out. Run that vvp output to test the simulation results. Collect
+    the results and loog for a "PASSED" string.'''
```
larsclausen, line 208:
```suggestion
    the results and look for a "PASSED" string.'''
```

ivtest/run_ivl.py
```diff
@@ -240,8 +199,99 @@ def do_run_normal(options : dict, expected_fail : bool) -> list:
         return [1, "Failed - No PASSED output, and no gold file"]
 
 
+def do_run_normal_vlog95(options : dict, expected_fail : bool) -> list:
+    '''Run the iverilog and vvp commands.
+
+    In this casem, run the compiler with the -tvlog95 flag to generate
```
larsclausen, line 205:
```suggestion
    In this case, run the compiler with the -tvlog95 flag to generate
```

PR #877, opened 2023-03-05
Handle continuous assignment of assignment patterns to array elements

larsclausen:
Currently when creating the NetNet for a continuous assignment to an array
element the type of the element is flattened into a canonical 1 dimensional
form. This works for most cases because packed types are compatible if
their total packed with is the same.

But there are some contexts such as if the right-hand-side is an assignment
pattern where the actual type matters and flattening the type will result
in incorrect behavior.

Retain the original type of the array element when creating the NetNet for
the array element assignment.

### DIFF
--- a/ivtest/regress-sv.list
+++ b/ivtest/regress-sv.list
@@ -521,11 +521,15 @@ struct_signed		normal,-g2009		ivltests
 sv-constants		normal,-g2005-sv	ivltests
 sv_ap_parray1		normal,-g2005-sv	ivltests
 sv_ap_parray2		normal,-g2005-sv	ivltests
+sv_ap_parray3		normal,-g2005-sv	ivltests
+sv_ap_parray4		normal,-g2005-sv	ivltests
 sv_ap_parray_fail1	CE,-g2005-sv		ivltests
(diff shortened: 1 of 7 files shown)


ivtest/regress-sv.list
```diff
@@ -521,11 +521,15 @@ struct_signed		normal,-g2009		ivltests
 sv-constants		normal,-g2005-sv	ivltests
 sv_ap_parray1		normal,-g2005-sv	ivltests
 sv_ap_parray2		normal,-g2005-sv	ivltests
+sv_ap_parray3		normal,-g2005-sv	ivltests
+sv_ap_parray4		normal,-g2005-sv	ivltests
```
steveicarus, line 525: Can you move these tests to the new format? Check out the regress-vvp.list, and the documentation here: https://steveicarus.github.io/iverilog/developer/regression_tests.html

I'm trying to close out the old format and move things forward. If the documentation is not clear, I'm open to suggestions. (Or PR's, of course.)

larsclausen: I was hoping we could just do automatic bulk conversion of all of these at once.

PR #880, opened 2023-03-08
Print more information about the SDF header

mole99:
This PR adds some printfs to output all the information about the SDF header.

I think this behavior is useful because this output is only printed on `-sdf-info` or `-sdf-verbose` and if you specify these arguments, you probably want all the information.

This PR adds the following output:

- SDF Version
- Divider
- Voltage (triple or scalar)
- Temperature (triple or scalar)

Which values are chosen from the min:typ:max expressions is determined with the `-Tmin|typ|max` argument passed to iverilog.

I tried to match the indentation style as it is, but it wasn't always clear to me what to choose. I think a unified indentation style, enforced by an auto formatter, would benefit Icarus very much. (But that's something for another time.)

I created a test (`sdf_header`) that reads in an sdf file with a complete header and compares the output to a golden reference. The test is registered in the new regression test script.

Since this is my first PR here, any comments regarding coding style/workflow/etc. are all very much welcome. Thanks!

### DIFF
--- a/ivtest/gold/sdf_header-vvp-stdout.gold
+++ b/ivtest/gold/sdf_header-vvp-stdout.gold
@@ -1,0 +1,10 @@
+ivltests/sdf_header.sdf:2:SDF INFO: Version: 3.0
+ivltests/sdf_header.sdf:3:SDF INFO: Design: test
+ivltests/sdf_header.sdf:4:SDF INFO: Date: Wed Mar 8 12:34:56 2023
+ivltests/sdf_header.sdf:5:SDF INFO: Vendor: No Vendor
+ivltests/sdf_header.sdf:6:SDF INFO: Program: Human
+ivltests/sdf_header.sdf:7:SDF INFO: Program Version: 1.0.0
+ivltests/sdf_header.sdf:9:SDF INFO: Voltage : 5.000000
+ivltests/sdf_header.sdf:10:SDF INFO: Process: best=0.65:nom=1.0:worst=1.8
+ivltests/sdf_header.sdf:11:SDF INFO: Temperature : 25.000000
+ivltests/sdf_header.sdf:12:SDF INFO: Timescale : 1ps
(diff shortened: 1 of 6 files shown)


mole99:
Rebased on master and removed empty gold files as per #881

mole99:
Rebased on master

svenka3:
I tried this today and observed the following:

```log
vvp   vsdf -sdf-verbose
sample.sdf:2:SDF INFO: Version: 3.0
sample.sdf:3:SDF INFO: Divider: "."
sample.sdf:4:SDF INFO: Timescale: 1ns
SDF WARNING: ../vlog_delay_mode_sdf.sv:42: Module dut in dut_inv_with_delays is not a in_w_del; it is a inv_w_del
```
Few notes/observations

1. If I provide "vvp -sdf-verbose vsdf" I get weird errors, perhaps known issue - is it easy enough to fix? If yes and given some pointers I can explore
2. On the "SDF INFO" I really like the output, just a nit-pick (sorry about it) - should we get "SDF INFO" first and then file-name:line information (a la "SDF WARNING: " - easy to fix, just running by authors.

Thanks

mole99:
Hi! Thanks for testing it.

1. With weird errors, do you mean your warning?: `SDF WARNING: ../vlog_delay_mode_sdf.sv:42: Module dut in dut_inv_with_delays is not a in_w_del; it is a inv_w_del`

If yes, I don't think that this PR changes anything in that regard. You could check if your `dut` is of type `inv_w_del` and the CELLTYPE in the SDF file is `in_w_del`. Then this warning would make sense.

2. Yes, I totally agree. `SDF INFO` should be at the start of the line to make it easier to see and to be consistent with `SDF WARNING`. I will open another PR in the next days 👍️

svenka3:
Thanks for quick response. That #1 was more of vvp use-model than this PR. I will investigate bit more. On this INFO, few more points that are generally useful are:

1. MTM - choice being done by user/tool
2. Instance on which the SDF is annotated (am not sure if there can be multiple etc. - but top instance would be good to have in such cases)

If you want, I can attempt those additional fixes, but since you offered to create PR, am sharing my ideas. 
Regards

mole99:
You are welcome. By the way, if you have an example for 1 I could take a look at it.

About INFO:

- What exactly do you mean with MTM?

- I think this is generally a good idea. There can be multiple instances, but I have to look into it a bit more.

Yes, I would just start with a simple PR to just move all `SDF INFO` at the start of the line and then I will see where I will go from there :)

PR #896, opened 2023-04-16
Added extra debug info - SDF annotation

### LINKED ISSUE #889: SDF warning/error enhancement proposal
I understand SDF support is limited and there is a GSOC project open as well. I came across a (very) low-hanging fruit type of enhancement and I have a local patch. Want to know if you are interested in a PR.

In my case, the DUT instance to be annotated is specified only inside SDF file, not during $sdf_annotate call (in Verilog). Wo when the instance name (or module name) is misspelt inside SDF I get:

```log
SDF WARNING: ../vlog_delay_mode_sdf.sv:42: Unable to find inst1 in scope dut_inv_with_delays. 
```
That line 42 in Verilog looks as follows:
```verilog
    $sdf_annotate("sample.sdf");  //, inst);
```
Here vvp found an instance named inst1 and that source is actually SDF, relevant line below:

```verilog
  (INSTANCE inst1)
```
So, a first-timer would be lost as to where the fix shall be. I enhanced it to pass SDF line number from sdf_parse.y to sys_sdf.c and have the below:

```log
SDF WARNING: ../vlog_delay_mode_sdf.sv:42: Unable to find inst1 in scope dut_inv_with_delays. SDF: sample.sdf 5
```
It is no rocket science, but does improve the usability IMHO. Also it is not perfect as it prints the line-number of CELL than the INSTANCE, but nevertheless it is showing the SDF file name and a close line. 

I do intend to enhance few more such low-hanging stuff around this area if this is of interest.

Thanks

svenka3:
Added extra debug info in case of SDF and Verilog dut/instance mismatches.

Fixes #889

### DIFF
--- a/vpi/sdf_parse.y
+++ b/vpi/sdf_parse.y
@@ -227,7 +227,7 @@ cell_list
 
 cell
   : '(' K_CELL celltype cell_instance
-      { sdf_select_instance($3, $4); /* find the instance in the design */}
+      { sdf_select_instance($3, $4, @2.first_line); /* find the instance in the design */}
     timing_spec_list_opt
     ')'
       { free($3);
--- a/vpi/sdf_priv.h
+++ b/vpi/sdf_priv.h
@@ -48,7 +48,7 @@ struct sdf_delval_list_s {
       struct sdf_delay_s val[12];
 };
 
-extern void sdf_select_instance(const char*celltype, const char*inst);
+extern void sdf_select_instance(const char*celltype, const char*inst, int sdf_line_num);
 extern void sdf_iopath_delays(int vpi_edge, const char*src, const char*dst,
 			      const struct sdf_delval_list_s*delval);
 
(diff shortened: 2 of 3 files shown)


caryr (changes requested):
This looks reasonable, but you are not using the appropriate spacing and there should be a test that checks the generated message.

caryr:
Also does it make more sense to only use the SDF file name and line numbers or include them at the beginning of the message.

caryr:
Maybe something like "...SDF_file:SDF_line loaded from SV_file:SV_line: ..." The wrong line number should be investigated. Maybe something is being ignored, but is not contributing to increasing the line count, etc.

caryr:
I have pushed changes that should add the file and line information you requested. It also adds this for a few cases you missed.

Note that the incorrect line number issues will need to be resolved separately since that will need a specific SDF file to reproduce or it may get fixed as part of the GSOC project.

PR #936, opened 2023-06-11
 Allow to omit trailing module ports in ordered list connection

### LINKED ISSUE #931: Missing connection for ordered list connection to an input port with default value result in error
I've got a module where the last port is an input with a default value specified.
When trying to instanciate it using ordered list connection and ignoring the last param it results in error :
`error: Wrong number of ports. Expecting 3, got 2`

However the LRM explicitely allows this (23.3.2.1) this is actually pretty much one of their example of things that should work and where the port should get the default value.

```verilog
module def(
    output logic a,
    input  logic x,
    input  logic y = 1'b1
);  
    assign a = x ^ y;
endmodule


module test(
    output logic a,
    input  logic x
);  
    def def_I ( a, x );
endmodule
```

Testing with `-g2012` flag on `Icarus Verilog version 13.0 (devel) (s20221226-127-gdeeac2ed-dirty)`

larsclausen:
The current implementation expects that for a module instantiation with a
ordered list connection all ports are supplied.

But there doesn't seem to be such a requirement in the LRMs. The Verilog
LRM doesn't mention anything in this regard and the SystemVerilog LRM
mentions in section 23.3.2.1 that a blank or omitted port connection is
either left unconnected or uses the default value of the port.

Update the implementation so that it allows to omit trailing ports and only
generates an error message if too many ports are specified in the ordered
port list.

Resolves #931

### DIFF
--- a/elaborate.cc
+++ b/elaborate.cc
@@ -1317,22 +1317,20 @@ void PGModule::elaborate_mod_(Design*des, Module*rmod, NetScope*scope) const
       } else {
 
 	      /* Otherwise, this is a positional list of port
-		 connections. In this case, the port count must be
-		 right. Check that is is, the get the pin list. */
+		 connections. Use as many ports as provided. Trailing
+		 missing ports will be left unconnect or use the default
+		 value if one is available */
 
-	    if (pin_count() != rmod->port_count()) {
+	    if (pin_count() > rmod->port_count()) {
 		  cerr << get_fileline() << ": error: Wrong number "
-			"of ports. Expecting " << rmod->port_count() <<
+			"of ports. Expecting at most " << rmod->port_count() <<
 			", got " << pin_count() << "."
 		       << endl;
 		  des->errors += 1;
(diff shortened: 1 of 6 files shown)


larsclausen:
I'd like to hear a second opinion on this interpretation of the LRM. In my opinion this makes the most sense, but I can also see how you'd get a slightly different interpretation with slightly different behavior.

caryr:
Does it make sense to print a warning if there is an unconnected port? If there is a default expression then it makes sense that is expected, but a truly unconnected port is likely an error in the code.

larsclausen:
It will print a warning with `-Wall` same way as if a port was missing for a named list connection.

caryr:
Okay, I'll look at the code and the standard later to see how this aligns with my interpretation.

caryr:
There is a bit more discussion in 23.3.3.2 and 23.3.3.3, but it is clear all ports except for ref ports can be left unconnected.

PR #942, opened 2023-06-14
Fix line location and require SystemVerilog mode for implicit named port connections

### LINKED ISSUE #939: Error reported at the wrong line
When running iverilog c74048a526a on this file 
```
module toto;
endmodule

module top;
        sub i(.o);
endmodule

module sub(output o);
endmodule
```
The error is reported at line 1 instead of 5:
```
test.v:1: error: Net o is not defined in this context.
test.v:1: error: Output port expression must support continuous assignment.
test.v:1:      : Port 1 (o) of sub is connected to o
```

larsclausen:
Two small fixes/changes for implicit named port connections.

- Use the correct line location
- Require SystemVerilog mode
  -  Regression test `br_gh315` is modified to run in SystemVerilog mode since it
makes use of implicit named port connections.

In order to regression test for the line information also add support for gold files for CE type tests  to vvp_reg.py.

Resolves #939

### DIFF
--- a/ivtest/gold/br_gh939-iverilog-stderr.gold
+++ b/ivtest/gold/br_gh939-iverilog-stderr.gold
@@ -1,0 +1,4 @@
+ivltests/br_gh939.v:10: error: Net o is not defined in this context.
+ivltests/br_gh939.v:10: error: Output port expression must support continuous assignment.
+ivltests/br_gh939.v:10:      : Port 1 (o) of M is connected to o
+2 error(s) during elaboration.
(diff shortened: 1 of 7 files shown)


caryr:
It looks like your gold file line number is out of date, it should be 10 not 8, and there is also one other failing VVP test.

larsclausen:
> It looks like your gold file line number is out of date, it should be 10 not 8, and there is also one other failing VVP test.

At least we know the gold file check is working :wink:. Fixed both issues now

PR #952, opened 2023-06-16
Make result of binary operations 2-state if inputs are 2-state

larsclausen:
The are many binary operations where if the two operands are 2-state the result is guaranteed to be 2-state.

This is true for all arithmetic operation with the exception of division where division by 0 will always result in 'x even if the inputs are both 2-state.

The same is true for all binary bitwise operators as well as the binary logical operators.

Having the expression type be 2-state avoids some unnecessary %cast2 instructions that would otherwise get inserted when assigning the result to a 2-state variable.

E.g without this change the following will result in

```
  int a, b, c;
  b = a + b;
```

will result in

```
  %load/vec4 ...;
  %load/vec4 ...;
  %add;
  %cast2;
  %store/vec4 ...;
```

For binary comparison operators this is already handled.

### DIFF
--- a/eval_tree.cc
+++ b/eval_tree.cc
@@ -807,7 +807,7 @@ NetEConst* NetEBLogic::eval_arguments_(const NetExpr*l, const NetExpr*r) const
 {
 	// NetEBLogic arguments should have already been reduced so real is not possible.
       ivl_assert(*this, (l->expr_type() != IVL_VT_REAL) && (r->expr_type() != IVL_VT_REAL));
-      ivl_assert(*this, expr_type() == IVL_VT_LOGIC);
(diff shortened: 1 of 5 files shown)


caryr:
For a shift I would expect the expression type to only depend on the left expression since the right expression is self-determined. I will check the standard later to confirm. Maybe it is only the width and not the type that works this way.

larsclausen:
The way I understand it is that any operation on 2-state types shall be equivalent to as if the operands were 4-state types. And then for certain types of operations it is known for sure that the result is within the 2-state type range and we can make the optimization to treat it that way. See section 11.3.4 in the TRM.

For shift that means if the right-hand-side is 'x or 'z the result will be 'x, even if the left-hand-side is 2-state.

PR #973, opened 2023-07-19
Implement SDF INTERCONNECT

mole99:
This PR implements the SDF INTERCONNECT feature and adds three tests to the test suite. While this base implementation can already correctly simulate interconnection delays for GL designs, some features are still missing for greater usability, namely:

- Support for interconnection delays of input/output vectors
- Reannotation of already annotated intermodpaths

I would prefer to add these features in a follow-up PR to keep this PR from growing too much. Note that there seems to be some issues with the CI as MINGW32 - i686 sometimes fails during the build process with the error:

```
mingw32-make: INTERNAL: Exiting with 1 jobserver tokens available; should be 3!
==> ERROR: A failure occurred in build().
```

I believe that this is not caused by the changes in this PR.

---

The changes in Icarus Verilog are the following:

SDF VPI Module:

- Implement `sdf_interconnect_delays()`
  - Find port handles in the correct scope
  - Get the intermodpath by calling `vpi_handle_multi(vpiInterModPath, port1_handle, port2_handle);`
  - Place delays via `vpi_put_delays(intermodpath, &delays);`

VVP:

- Implement `__vpiInterModPath` for VPI access
- Implement `vvp_fun_intermodpath` as net functor
- Implement `vpi_handle_multi()` to return a `__vpiInterModPath`
  - Find nets for given ports
  - Create a `vvp_fun_intermodpath` and `__vpiInterModPath`
  - Insert intermodpath into netlist
  - return `__vpiInterModPath`

Three tests were created to exercise the new feature:

- sdf_interconnect1.v - A design with three buffers in series
- sdf_interconnect2.v - A design with three buffers in parallel
- sdf_interconnect3.v - A design with various buffers and xor gates

Designs 1 and 2 are trivial and can be checked by hand. Design 3 was checked by hand, but also compared with the CVC simulator and checked via my script in [this repository](https://github.com/mole99/interconnect-tests).

~As soon as #989 is merged I will update the tests to also include IOPATH delays.~ Done

### DIFF
--- a/compiler.h
+++ b/compiler.h
@@ -177,6 +177,11 @@ extern bool gn_icarus_misc_flag;
    is false, then skip elaboration of specify behavior. */
 extern bool gn_specify_blocks_flag;
 
+/* If this flag is true, then add input/output buffers to modules so that
+   VVP can insert intermodpaths inbetween. If this flag
+   is false, then no input/output buffers are inserted if not needed. */
+extern bool gn_interconnect_flag;
+
 /* If this flag is true, then elaborate supported assertion statements. If
    this flag is false, then stub out supported assertion statements. */
 extern bool gn_supported_assertions_flag;
--- a/driver/iverilog.man.in
+++ b/driver/iverilog.man.in
@@ -95,6 +95,14 @@ simulation, and in fact can hurt performance of the
 simulation. However, disabling specify blocks reduces accuracy of
 full-timing simulations.
 .TP 8
+.B -ginterconnect\fI|\fP-gno-interconnect
+Enable or disable (default) SDF interconnect support. When enabled,
+input and output buffers will be added to modules, so that VVP can
+add interconnection delays. This option is commonly not needed for RTL
+simulation, and in fact can hurt performance of the
+simulation. However, disabling interconnection delays reduces accuracy of
+full-timing simulations.
+.TP 8
 .B -gstd-include\fI|\fP-gno-std-include
 Enable (default) or disable the search of a standard installation
 include directory after all other explicit include directories. This
--- a/driver/main.c
+++ b/driver/main.c
@@ -125,6 +125,7 @@ char depmode = 'a';
 
 const char*generation = "2005";
 const char*gen_specify = "no-specify";
+const char*gen_interconnect = "no-interconnect";
 const char*gen_assertions = "assertions";
 const char*gen_xtypes = "xtypes";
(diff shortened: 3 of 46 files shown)


caryr:
Also there is some inconsistent spacing.

mole99:
Yes, the CI is really helpful :)

The issue was not with `vvp.def`, I had added the routine there, but I also had to add it to `libvpi.c` and `vpi_modules.cc` and update `vpip_routines_s`. 

Now it seems to work!

mole99:
Yes, my IDE does not handle this type of formatting well. I will try to correct the spacing during development and do a double check once the PR is ready.

caryr:
Much of this needs to have the formatting updated as discussed in your previous pull request.

mole99:
Yes, I will work on the formatting after my break next week.

mole99:
The formatting should now be cleaned up. Just tell me if you see something I missed so I can fix it.

caryr:
I realize this is slightly maddening, but the formatting is different in parse files. The indenting for the code in the first `{}` pair is different since this is used to delineate the C/C++ from the parse actions.

Also having different values for the min/typ/max in the SDF files allows you to find cases where the wrong value is being used. The scheme I had used previously allows you to easily determine if no delay, the min, typ or max delay was being annotated.

caryr:
For an interconnect port I believe a value of -1 can be valid if the port is declared to contain that index. I realize most ports are zero based, but we cannot assume that is the case. I think there are cases where the minimum value is used, but -1 seems a bit risky as the indicator for no index value being provided (sdf_parse.y)

mole99:
Thanks, I did not think of negative indices. I extended `interconnect_port_s` with the flag `has_index` which indicates that the index for the bit of a vector is in use.

Regarding the formatting, how is the formatting of the first `{}` pair different? I am unsure about how to change it. Is it because once Bison is run, the code is inside a `case` statement, such as here?

grafik


What scheme are you referring to for the min/typ/max values? I had a look at the other tests and most of the SDF files define the same value for min/typ/max.

caryr:
I thought for some I used `1:2:3` or some other appropriate scaled version.

```
{ first line
  second line
  third line
}
```
This is what I was talking about for the SDF parse files. If you look at the code you should find this is a common pattern. I'm going to push this for now and you can clean things in a later pull request.

mole99:
Thanks, I will correct the formatting in the follow up PR.

PR #989, opened 2023-08-14
Always evaluate `vvp_fun_modpath_src` first

### LINKED ISSUE #985: Issue with modpaths when BUFT is in different scope
This is a very specify issue that only occurs when manually editing the VVP source file, but an issue nonetheless.

For my work on SDF interconnect, I need to add input and output buffers to modules. Below is the VVP code for a simple design, a buffer `buffer0` instantiated inside the toplevel module `top`. Instead of `a` and `b` being directly connected to the `BUFZ` of `buffer0`, both signals go through a `BUFT`.

```
#! /usr/local/bin/vvp
:ivl_version "13.0 (devel)" "(s20221226-244-g3e96a7e1e)";
:ivl_delay_selection "TYPICAL";
:vpi_time_precision - 12;
:vpi_module "/usr/local/lib/ivl/system.vpi";
:vpi_module "/usr/local/lib/ivl/vhdl_sys.vpi";
:vpi_module "/usr/local/lib/ivl/vhdl_textio.vpi";
:vpi_module "/usr/local/lib/ivl/v2005_math.vpi";
:vpi_module "/usr/local/lib/ivl/va_math.vpi";
S_0x5650ecc5e540 .scope module, "top" "top" 2 15;
 .timescale -9 -12;
v0x5650ecc95a40_0 .var "a", 0 0;
v0x5650ecc95b00_0 .net "b", 0 0, L_0x5650ecc95d40;  1 drivers
S_0x5650ecc85830 .scope module, "buffer0" "buffer" 2 26, 2 3 0, S_0x5650ecc5e540;
 .timescale -9 -12;
    .port_info 0 /INPUT 1 "in";
    .port_info 1 /OUTPUT 1 "out";
L_0x5650ecc95bc0 .functor BUFZ 1, L_0x5650ecc95c80, C4<0>, C4<0>, C4<0>;
L_0x5650ecc95c80 .functor BUFT 1, v0x5650ecc95a40_0, C4<0>, C4<0>, C4<0>;
L_0x5650ecc95d40 .functor BUFT 1, V_0x5650ecc95920_0/m, C4<0>, C4<0>, C4<0>;
v0x5650ecc85a60_0 .net "in", 0 0, L_0x5650ecc95c80;  1 drivers
v0x5650ecc95920_0 .net "out", 0 0, V_0x5650ecc95920_0/m;  1 drivers
  .scope S_0x5650ecc85830;
V_0x5650ecc95920_0/m .modpath 1 L_0x5650ecc95bc0 v0x5650ecc95920_0,
   L_0x5650ecc95c80 (1000,1000,1000, 1000,1000,1000, 1000,1000,1000, 1000,1000,1000) v0x5650ecc85a60_0;
    .scope S_0x5650ecc5e540;
T_0 ;
    %vpi_call 2 19 "$dumpfile", "top.vcd" {0 0 0};
    %vpi_call 2 20 "$dumpvars", 32'sb00000000000000000000000000000000, S_0x5650ecc5e540 {0 0 0};
    %end;
    .thread T_0;
    .scope S_0x5650ecc5e540;
T_1 ;
    %delay 5000, 0;
    %pushi/vec4 0, 0, 1;
    %assign/vec4 v0x5650ecc95a40_0, 0;
    %delay 10000, 0;
    %pushi/vec4 1, 0, 1;
    %assign/vec4 v0x5650ecc95a40_0, 0;
    %delay 10000, 0;
    %vpi_call 2 37 "$finish" {0 0 0};
    %end;
    .thread T_1;
# The file index is used to find the file name in the following table.
:file_names 3;
    "N/A";
    "<interactive>";
    "top.v";
```

The expected result is that b is delayed by one ns. However, if you run this VVP source file with the latest master, the delay inside the specify block is not applied.

No delay

If the "BUFTs" are manually moved to the outer scope `top`, the simulation behaves correctly.

Correct delay

For example, insert the `BUFT`s directly after `a` and `b`.

```
v0x5650ecc95a40_0 .var "a", 0 0;
v0x5650ecc95b00_0 .net "b", 0 0, L_0x5650ecc95d40;  1 drivers
L_0x5650ecc95c80 .functor BUFT 1, v0x5650ecc95a40_0, C4<0>, C4<0>, C4<0>;
L_0x5650ecc95d40 .functor BUFT 1, V_0x5650ecc95920_0/m, C4<0>, C4<0>, C4<0>;
```

This is unexpected because the netlist stays the same, only the scope of the `BUFT`s has changed.

I am going to investigate this problem, because for SDF interconnect work I need input and output buffers in the same scope as the `.port_info` directives, so I can add a reference to the associated buffer.

But if someone already has a clue what it could be, please tell me :)

mole99:
Fixes #985

This PR ensures that whenever any `vvp_net_t`s are connected via `vvp_net_t::link`, nodes with `vvp_fun_modpath_src` are always inserted at the head of the linked list.

This guarantees that `vvp_fun_modpath_src`  are always evaluated before their respective `vvp_fun_modpath`, without any runtime penalty.

### DIFF
--- a/vvp/vvp_net.cc
+++ b/vvp/vvp_net.cc
@@ -199,8 +199,38 @@ vvp_net_t::vvp_net_t()
 void vvp_net_t::link(vvp_net_ptr_t port_to_link)
 {
       vvp_net_t*net = port_to_link.ptr();
-      net->port[port_to_link.port()] = out_;
-      out_ = port_to_link;
+
+      // Connect nodes with vvp_fun_modpath_src to the head of the
+      // linked list, so that vvp_fun_modpath_src are always evaluated
+      // before their respective vvp_fun_modpath
+      if (dynamic_cast<vvp_fun_modpath_src*>(net->fun))
+      {
+	      net->port[port_to_link.port()] = out_;
+	      out_ = port_to_link;
+      }
+      // All other nodes connect to the tail of the linked list
+      else
+      {
+	      // List is empty, just point to port_to_link
+	      if (out_.ptr() == nullptr)
+	      {
+		      out_ = port_to_link;
+		      net->port[port_to_link.port()] = vvp_net_ptr_t(nullptr,0);
+		      return;
+	      }
+
+	      vvp_net_ptr_t cur = out_;
+
+	      // Traverse through linked list
+	      while (cur.ptr()->port[cur.port()].ptr())
+	      {
+		      cur = cur.ptr()->port[cur.port()];
+	      }
+
+	      // Connect to the tail of list
+	      cur.ptr()->port[cur.port()] = port_to_link;
+	      net->port[port_to_link.port()] = vvp_net_ptr_t(nullptr,0);
+      }
 }
 
 /*
--- a/vvp/vvp_net.cc
+++ b/vvp/vvp_net.cc
@@ -208,28 +208,34 @@ void vvp_net_t::link(vvp_net_ptr_t port_to_link)
 	      net->port[port_to_link.port()] = out_;
 	      out_ = port_to_link;
       }
-      // All other nodes connect to the tail of the linked list
+      // All other nodes connect after the last vvp_fun_modpath_src
       else
       {
-	      // List is empty, just point to port_to_link
-	      if (out_.ptr() == nullptr)
+	      // List is empty or first node has no vvp_fun_modpath_src
+	      // then just insert the node into the head of linked list
+	      if (out_.ptr() == nullptr || ! dynamic_cast<vvp_fun_modpath_src*>(out_.ptr()->fun))
 	      {
+		      net->port[port_to_link.port()] = out_;
 		      out_ = port_to_link;
-		      net->port[port_to_link.port()] = vvp_net_ptr_t(nullptr,0);
 		      return;
 	      }
 
 	      vvp_net_ptr_t cur = out_;
+	      vvp_net_ptr_t prev = vvp_net_ptr_t(nullptr,0);
 
 	      // Traverse through linked list
-	      while (cur.ptr()->port[cur.port()].ptr())
+	      while (dynamic_cast<vvp_fun_modpath_src*>(cur.ptr()->fun))
 	      {
+		      prev = cur;
 		      cur = cur.ptr()->port[cur.port()];
+		      if (!cur.ptr()) break; // End of list
 	      }
 
-	      // Connect to the tail of list
-	      cur.ptr()->port[cur.port()] = port_to_link;
-	      net->port[port_to_link.port()] = vvp_net_ptr_t(nullptr,0);
+	      assert(prev.ptr());
+
+	      // Insert after last vvp_fun_modpath_src (or end of list)
+	      net->port[port_to_link.port()] = cur;
+	      prev.ptr()->port[prev.port()] = port_to_link;
       }
 }
 


mole99:
3 test cases have failed:
- pr1002
- vcd-dup
- partselsynth


This is surprising since the change in this PR should do nothing but change the order of the nodes in the linked list. The resulting netlist is the same. I also don't see any connection between these test cases.

@caryr, do you by any chance have an idea why these test cases fail?

caryr:
Do they fail locally? The log directory in the test directory usually has the details why a given test fails.

If you are not running the local tests you should be doing that as it gives you a good idea regarding the health of your code before you commit.

mole99:
Yes they do fail locally, but I do not see anything obvious in the resulting log files, like for example a forgotten debug statement.

The gold file for `pr1002` contains:

```
datain =  4095 dataout =   1023 expected =   1023 ... CHECK PASSED
datain =     0 dataout =      0 expected =      0 ... CHECK PASSED
datain =  8191 dataout =   2047 expected =   2047 ... CHECK PASSED
datain =  4096 dataout =   1024 expected =   1024 ... CHECK PASSED
datain = -4097 dataout =  -1025 expected =  -1025 ... CHECK PASSED
datain = -8192 dataout =  -2048 expected =  -2048 ... CHECK PASSED
datain =    -1 dataout =     -1 expected =     -1 ... CHECK PASSED
datain = -4096 dataout =  -1024 expected =  -1024 ... CHECK PASSED
TEST PASSED :-)
```

And the log file:

```
datain =  4095 dataout =   1023 expected =      x ... CHECK PASSED
datain =     0 dataout = 0000 expected = 03ff ... CHECK FAILED
datain =  8191 dataout = 07ff expected = 0000 ... CHECK FAILED
datain =  4096 dataout = 0400 expected = 07ff ... CHECK FAILED
datain = -4097 dataout = fbff expected = 0400 ... CHECK FAILED
datain = -8192 dataout = f800 expected = fbff ... CHECK FAILED
datain =    -1 dataout = ffff expected = f800 ... CHECK FAILED
datain = -4096 dataout = fc00 expected = ffff ... CHECK FAILED
TEST FAILED :-(
```

The gold file for `vcd-dup` contains:

```
VCD info: dumpfile work/vcd-dup.vcd opened for output.
VCD warning: skipping signal test.m2.c1, it was previously included.
VCD warning: ignoring signals in previously scanned scope test.m1.
VCD warning: $dumpvars ignored, previously called at simtime 0
```

And the log file:

```
VCD info: dumpfile work/vcd-dup.vcd opened for output.
VCD warning: skipping signal test.m2.c1, it was previously included.
VCD warning: ignoring signals in previously scanned scope test.m1.
VCD warning: $dumpvars ignored, previously called at simtime 0
./ivltests/vcd-dup.v:49: $finish called at 28 (1s)
```

Also a `.vcd` file is generated that is different to the gold `.vcd` file. I wonder how you even properly compare `.vcd` files because the `$date` and the naming of the variables changes.

Lastly the log file for `partselsynth` (no gold file):

```
./ivltests/partselsynth.v:8: warning: Process not synthesized.
FAILED -- scon=00000000, enable=0, bar_a==1, bar_b=0, val=0
./ivltests/partselsynth.v:56: $finish called at 5 (1s)
```

I don't know how my changes could lead to failures in such seemingly unrelated test cases.

caryr:
You can also look at the partselsynth test failure since that is giving a specific value combination that is failing.

mole99:
Yes there is a change in evaluation order, because previously nodes were always inserted at the head of the list, now all other nodes except those with `vvp_fun_modpath_src` are inserted at the tail of the list, therefore the order for all other nodes is reversed. I thought this would make no difference as `vvp_fun_modpath_src` is the only functor where the order matters. But that turned out to be wrong :)

I tried to keep the order of all other nodes by always inserting them right after the last `vvp_fun_modpath_src`, this way the order for all other nodes is the same. Now all test cases are successful.

Thanks for your help!

caryr:
The indenting and code formatting does not match what we use in the rest of the code.We try to keep this consistent so people do not get confused with the style of the code. Can you fix this in all your pull requests?

caryr:
Would it make sense to use a different data structure so you do not need to traverse the list every time you want to add something at the end? Maybe the total items added is small enough this is never an issue, but have you considered this? I believe this is part of the build stage so a minor increase is probably okay.

mole99:
For formatting, I tried to follow the general rule of six spaces or a combination of tabs and spaces to give a multiple of six. The function `vvp_net_t::unlink` seems to use six spaces as the first indentation level, and one tab and four spaces for the second indentation level. The function below that, `vvp_net_t::count_drivers`, uses only spaces.

I am confused about how to format my code since the existing code base contradicts itself. I feel like iverilog would benefit a lot from a code-formatter, maybe I'll come back to it after GSoC.

But for now, what formatting should I use for `vvp_net_t::link`?

mole99:
I don't think it makes sense to use a different data structure now. As you have stated, this is only part of the build stage, and the fanout for signals is usually small. Also, the impact of a different data structure on other parts of the simulator cannot be predicted. As we have seen some features for some tests rely heavily on the internal representation of the netlist.

caryr:
It's probably easier for me to make the changes so I'll merge this and you can look at my commit later.

mole99:
Thanks, I have looked at how you formatted the code. It seems like the rule here is to one tab for comments, 6 spaces for one indentation level and one tab and 4 spaces for the second indentation level.

I have now rebased PR #973 onto master, if you find some time it would be ready for review.

caryr:
It's all effectively 6 spaces for the code, 6 spaces, a tab and 4 spaces (The tab is assumed to be 8 spaces), two tabs and two spaces, three tabs, etc. The comments are normally indented two spaces relative to the relevant code, but use tabs as necessary. Case statements, classes and others are a little different since some items (e.g. case labels) have less indent than expected, but the major enclosed items follow the spacing rule. You were also not following the conventional block rules for example:
```
if (<condition>) {
      <true code>
} else {
      <false code>
}
```
is how we would write this not:
```
if (<condition>)
{
      <true code>
}
else
{
      <false code>
}
```
Line continuations are also tricky. I try to give them the same spacing as the first line and then fill the rest with spaces since that keeps the lines aligned exactly as expected even when the tab could hide initial characters (e.g. git diff). For example:
```
<tab>  var = <first part of a long expr> &&
<tab>        <second part of a long expr>;
```
So for this I like to split at an operator and leave the operator at the end of the line being continued so you have a strong clue the line is being continued. Yes not having a ";" should indicate this, but it is easy to miss. The same should apply to any line continuation where the continued part is aligned to the corresponding  code from the previous line and only uses spaces to make up the difference from the base line indent. Here's a more complicated example:
```
<tab>  var = <long guard expr> && (<first part of a long expr> ||
<tab>                              <second part of a long expr>);
```
This is almost required to keep nested ternary operators straight and I often continue these just for clarity:
```
val = <first cond> ?  <second cond> ? <second true> :
                                      <second false> :
                      <first false>;
```
Probably more things, but this covers most of what I have seen in your code. I suppose all this should go into a document somewhere. I'll try to work on that.

I plan to review this next week. I have a major deadline at work this week and some personal things to take care of as well. So am already overly committed.

mole99:
Thanks for the write up on the formatting, it makes things clearer.

Please take your time with the review. I will simply continue with my next work ontop of this PR, so it is not urgent. I really appreciate the time you have already invested as a mentor, especially since you are practically the only mentor capable of helping me with iverilogs codebase :)

PR #992, opened 2023-08-19
Add support for binding function/task arguments by name

### LINKED ISSUE #233: SystemVerilog: Function call by named association not supported
SystemVerilog allows the usage of function calls with named association to pass arguments. Icarus generates an invalid syntax error when this is used.
I'm using iverilog 10.1 (stable).

Example code:
```systemverilog
package TestPkg;

    function automatic logic [31:0] encode(
        logic [11:0] a,
        logic [4:0] b,
        logic [2:0] c,
        logic [4:0] d,
        logic [6:0] e
    );
        begin
            encode[31:20] = a;
            encode[19:15] = b;
            encode[14:12] = c;
            encode[11:7] = d;
            encode[6:0] = e;
        end
    endfunction

endpackage
```
Module that works, compiled with `iverilog -Wall -tnull -g2012 TestPkg.sv FuncUnnamedAssoc.sv` :
```systemverilog
module FuncUnnamedAssoc(
    output logic [31:0] encoded,
    input logic [11:0] f,
    input logic [4:0] g,
    input logic [2:0] h,
    input logic [4:0] i,
    input logic [6:0] j
);

import TestPkg::*;

always @(*) begin
    encoded = encode(
        f,
        g,
        h,
        i,
        j
    );
end
```
Module that generates a syntax error, compiled with `iverilog -Wall -tnull -g2012 TestPkg.sv FuncNamedAssoc.sv` :
```systemverilog
module FuncNamedAssoc(
    output logic [31:0] encoded,
    input logic [11:0] f,
    input logic [4:0] g,
    input logic [2:0] h,
    input logic [4:0] i,
    input logic [6:0] j
);

import TestPkg::*;

always @(*) begin
    encoded = encode(
        .a(f),
        .b(g),
        .c(h),
        .d(i),
        .e(j)
    );
end

endmodule
```
Error message is:
FuncNamedAssoc.sv:14: syntax error
FuncNamedAssoc.sv:19: error: malformed statement

larsclausen:
In addition to providing positional arguments for task and functions
SystemVerilog allows to bind arguments by name. This is similar to how
module ports can be bound by name.

```SystemVerilog
task t(int a, int b); ... endtask
...
t(.b(1), .a(2));
```

Extend the parser and elaboration stage to be able to handle this. During
elaboration the named argument list is transformed into a purely positional
list so that later stages like synthesis do not have to care about the
names.

For system functions and tasks all arguments must be unnamed, otherwise an
error will be reported.

In addition to functions and tasks arguments can also be bound by name for
the various different ways of invoking a class constructor.

Resolves #233

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -114,7 +114,8 @@ O = main.o async.o design_dump.o discipline.o dup_expr.o elaborate.o \
     net_event.o net_expr.o net_func.o \
     net_func_eval.o net_link.o net_modulo.o \
     net_nex_input.o net_nex_output.o net_proc.o net_scope.o net_tran.o \
-    net_udp.o pad_to_width.o parse.o parse_misc.o pform.o pform_analog.o \
+    net_udp.o map_named_args.o \
+    pad_to_width.o parse.o parse_misc.o pform.o pform_analog.o \
     pform_disciplines.o pform_dump.o pform_package.o pform_pclass.o \
     pform_types.o \
     symbol_search.o sync.o sys_funcs.o verinum.o verireal.o vpi_modules.o target.o \
--- a/PExpr.cc
+++ b/PExpr.cc
@@ -215,12 +215,12 @@ PEBShift::~PEBShift()
 {
 }
 
-PECallFunction::PECallFunction(const pform_name_t&n, const vector<PExpr *> &parms)
+PECallFunction::PECallFunction(const pform_name_t &n, const vector<named_pexpr_t> &parms)
 : path_(n), parms_(parms), is_overridden_(false)
 {
 }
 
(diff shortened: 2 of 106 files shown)


caryr:
I'm not certain if it makes a difference, but I noticed the two new files do not have an associated name or date in the copyright.

larsclausen:
> I'm not certain if it makes a difference, but I noticed the two new files do not have an associated name or date in the copyright.

I'm not sure either, I don't think it is strictly needed. But maybe its best we follow https://reuse.software/spec/#copyright-and-licensing-information for both the copyright and license information for new files. I've updated the files with that.

PR #993, opened 2023-08-29
add double quote escapes from pragma comments.

ithinuel:
This was noticed while using Vivado which does the expansion in pragma comments.

I am not sure the test is the right place to add it. Let me know if I need to update this.


caryr:
I am going to merge this, but there are a couple issues that I will fix so look at my changes if you are interested.

ithinuel:
Thank you!! :D

PR #994, opened 2023-08-30
SDF: Interconnection delays for input/output vectors

mole99:
This PR add support for annotation of interconnection delays for input/output vectors. Previously, only single bit inputs/outputs could be annotated with an interconnection delay.

- Adds a testcase with a 3 bit input and 3 bit output vector

Note: This PR depends on #973, I will rebase once it is merged.


caryr:
This pull request is still marked as a draft so let me know when you want me to review and commit these changs.

mole99:
The formatting should now also be correct.

mole99:
Yes, I too think a cleanup of the SDF messages is due. Let's check if we agree:

Error: Fatal, ends the current annotation of IOPATH, INTERCONNECT, TIMINGCHECK. (Should it also stop the simulator?)
Warning: Minor issue but annotation should continue if possible. (Unsupported/ignored constructs, multiple matching ports, etc.)
Info: Useful information for the user and for debugging.

I would also propose to enable warnings by default. The user may not be aware that warnings need to be enabled and this could lead to seemingly correct simulations.

Error: Always enabled
Warning: Enabled by default, disabled via `-sdf-no-warn`
Info: Disabled by default, enabled via `-sdf-info`

`-sdf-verbose` would then be removed

caryr:
We could have a `fatal` category that stops the simulator and an `error` category that aborts the annotation. Neither could be disabled. Add to your existing or start a new discussion so others can comment.

PR #999, opened 2023-09-12
Implement $width timing check (only vvp for now)

mole99:
This PR contains my latest work on implementing the `$width` timing check. It checks that the pulse width of a signal does not fall below a limit.

It is described by a record of type `.tchk_width`, e.g.:

```
.tchk_width 3 8, "posedge", v0x55a707f6d570_0, v0x000000000000_0, 1.0, 0, v0x55a707f6d450_0;
```

The individual elements:

    .tchk_width <file_idx> <line_num>, "<starting_edge>", <reference_signal>, <condition>, <limit>, <threshold>, <notifier_signal>;

Note: The automatic insertion of such a record into the `.vvp` file is not yet implemented, for now this has been tested manually.

The following is supported:

- Both "posedge" and "negedge" as starting edge
- Violation message
- SDF annotation
- cbTchkViolation VPI callback

The following is not supported:

- Conditions

I have got some tests prepared, but I will commit them only after the record insertion works.

Since my GSoC project is coming to an end and the next few weeks will be busy for me, it may be some time before I get back to finish this, but I do intend to.

However, what has already been implemented could be reviewed in the meantime, thanks!

### DIFF
--- a/vpi/sdf_lexor.lex
+++ b/vpi/sdf_lexor.lex
@@ -45,8 +45,6 @@ static int yywrap(void)
 %}
 
 %x CCOMMENT
-%x COND_EDGE_ID
-%x EDGE_ID
 
 %%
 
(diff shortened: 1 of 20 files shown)


mole99:
The CI fail is because the output for one testcase is now different.

mole99:
Unfortunately nothing has progressed here. I hope to find more time in the new year.

Issue #1012, opened 2023-10-20
vvp: internal error when concatenating verilog array elements


proppy:
The following verilog:
```
module ArraySliceWithNarrowStart(
  input wire [159:0] a,
  input wire start,
  output wire [95:0] out
);
  wire [31:0] a_unflattened[0:4];
  assign a_unflattened[0] = a[31:0];
  assign a_unflattened[1] = a[63:32];
  assign a_unflattened[2] = a[95:64];
  assign a_unflattened[3] = a[127:96];
  assign a_unflattened[4] = a[159:128];
  wire [31:0] array_slice_6[0:2];
  assign array_slice_6[0] = a_unflattened[{2'h0, start} > 3'h4 ? 3'h4 : {2'h0, start} + 3'h0];
  assign array_slice_6[1] = a_unflattened[{2'h0, start} > 3'h3 ? 3'h4 : {2'h0, start} + 3'h1];
  assign array_slice_6[2] = a_unflattened[{2'h0, start} > 3'h2 ? 3'h4 : {2'h0, start} + 3'h2];
  assign out = {array_slice_6[2], array_slice_6[1], array_slice_6[0]};
endmodule
```
triggers the following error in vvp:
```
internal error: port 0 expects wid=32, got wid=0
vvp: concat.cc:54: virtual void vvp_fun_concat::recv_vec4(vvp_net_ptr_t, const vvp_vector4_t&, vvp_context_t): Assertion `0' failed.
```

proppy:
managed to reduce the test case to the following:
```
module ArraySliceConcat(
  input wire start,  		      
  output wire [32:0] out
);
  wire [31:0] a_unflattened[0:4];
  assign out = a_unflattened[{2'h0, start} > 3'h4 ? 3'h0 : 3'h0];
endmodule
```

proppy:
seems to be related to the conditional indexing evaluation: `a_unflattened[{2'h0, start} > 3'h4 ? 3'h0 : 3'h0]`.

proppy:
fyi, it doesn't seem to fail with https://github.com/steveicarus/iverilog/tree/v11_0

proppy:
nor with with https://github.com/steveicarus/iverilog/tree/v12_0

larsclausen:
https://github.com/steveicarus/iverilog/commit/a7883aaf7d55b579514bda4509f5103a28402665 should fix it. Will create a PR later.

proppy:
[...]

PR #1014, opened 2023-10-23
ivtest: add array_slice_concat

proppy:
Related #1012

### DIFF
--- a/ivtest/ivltests/array_slice_concat.v
+++ b/ivtest/ivltests/array_slice_concat.v
@@ -1,0 +1,17 @@
+module ArraySliceWithNarrowStart(
+  input wire [159:0] a,
+  input wire start,
+  output wire [95:0] out
+);
+  wire [31:0] a_unflattened[0:4];
+  assign a_unflattened[0] = a[31:0];
+  assign a_unflattened[1] = a[63:32];
+  assign a_unflattened[2] = a[95:64];
+  assign a_unflattened[3] = a[127:96];
+  assign a_unflattened[4] = a[159:128];
+  wire [31:0] array_slice_6[0:2];
+  assign array_slice_6[0] = a_unflattened[{2'h0, start} > 3'h4 ? 3'h4 : {2'h0, start} + 3'h0];
+  assign array_slice_6[1] = a_unflattened[{2'h0, start} > 3'h3 ? 3'h4 : {2'h0, start} + 3'h1];
+  assign array_slice_6[2] = a_unflattened[{2'h0, start} > 3'h2 ? 3'h4 : {2'h0, start} + 3'h2];
+  assign out = {array_slice_6[2], array_slice_6[1], array_slice_6[0]};
+endmodule
--- a/ivtest/ivltests/array_slice_concat.v
+++ b/ivtest/ivltests/array_slice_concat.v
@@ -15,3 +15,21 @@ module ArraySliceWithNarrowStart(
   assign array_slice_6[2] = a_unflattened[{2'h0, start} > 3'h2 ? 3'h4 : {2'h0, start} + 3'h2];
   assign out = {array_slice_6[2], array_slice_6[1], array_slice_6[0]};
 endmodule
+
+module top;
+  reg [159:0] a;
+  reg start;
+  wire [95:0] out;
+  
+  ArraySliceWithNarrowStart dut(.a(a), .start(start), .out(out));
+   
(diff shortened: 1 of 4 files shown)


caryr:
Thank you for the example. Our normal tests check parsing and that the correct results is also generated. This does not have any results checking or indication of success. This could be marked CO (compile only) or as a minimum you could add an initial block that prints "PASSED" to indicate success. Printing nothing should be an indication of failure, but I need to double check that is the case.

proppy:
@caryr added a tb with 7bba9dd, but now I'm getting a different failure than #1012, even with @larsclausen's https://github.com/steveicarus/iverilog/issues/1012#issuecomment-1773944425 applied.
```
vvp: array.cc:166: virtual int __vpiArray::get_word_size() const: Assertion `vsig' failed.
```

steveicarus:
Please use the new format for all new [ivtest](https://steveicarus.github.io/iverilog/developer/regression_tests.html) submissions. In particular, do not touch any of the .list files, and instead create a vvp_tests/foo.json file that describes how to run your test.

proppy:
> Please use the new format for all new [ivtest](https://steveicarus.github.io/iverilog/developer/regression_tests.html) submissions. In particular, do not touch any of the .list files, and instead create a vvp_tests/foo.json file that describes how to run your test.

Done with 87965e3, PTAL.

caryr:
There was a change in the proposed fix. Is your test failing with that change or only the initial one?

steveicarus:
You created the vvp_tests/array_slice_concat.json file, and it looks fine, but you need to add it to the ivtest/regress-vvp.list file.

proppy:
@steveicarus noted! should be fixed with f9fb19c

proppy:
looks like the CI failure comes from https://github.com/steveicarus/iverilog/commit/56c5bf1da1d4540978a4db711b31c30937f26abf#diff-1d54053eed91417946beb50efd2a60e3af269b32af6425c12bbcc55486da8128R210-R213 (and a left over `expected_fail`).

steveicarus:
A fix has been merged. Try rebasing off the current master so that it can get through CI.

proppy:
> A fix has been merged. Try rebasing off the current master so that it can get through CI.

rebased, thanks!

larsclausen:
I've merged the fix. Sorry, you have to rebase one more time to make the tests pass.

proppy:
> I've merged the fix. Sorry, you have to rebase one more time to make the tests pass.

Done!

ivtest/ivltests/array_slice_concat.v
```diff
@@ -0,0 +1,35 @@
...
+  reg [159:0] a;
+  reg start;
+  wire [95:0] out;
+  
+  ArraySliceWithNarrowStart dut(.a(a), .start(start), .out(out));
+   
+  initial begin
+     a = {32'h44444444, 32'h33333333, 32'h22222222, 32'h11111111};
+     start = 1;
+     if (out !== 96'h444444443333333322222222) begin
```
larsclausen, line 29: Looks like the test is failing in CI. Might need a `#1` before the `if`.

proppy: It should be combinational (so no delay involved).

The logs say:
```
2023-12-30T02:13:41.5806676Z           array_slice_concat: Failed - No PASSED output, and no gold file
```
do I need to create a gold file?

caryr: No you do not need a gold file, but the test is also not passing as expected for the reason Lars mentioned. When the code is run it produces:

FAILED
tmp.v:31: $finish called at 0 (1s)

The reason for this has to do with the non-determinism in how the simulator handles events.

You are expecting that once `a` and `start` have been assigned values the value of `out` is immediately updated so that the if can verify the result is correct, but that is not how logic simulators are specified to work. Even no delay combinational logic has to be evaluated at a given time step and this does not happen immediately. What is happening in your code is that you are actually checking the value of `out` before it has been updated and that is the reason it is report this as a failing test.

It is always best to run your test code standalone to make sure it is running properly before submitting the change.

To make things work we need to stop the execution of the `initial begin/end` block so that the new output value can be calculated and that is done with some kind of delay or wait statement. In this case adding a `#1` after the variable assignments and before the `if` will allow the simulator to calculate the output and then the newly calculated value will be compared in the `if` statement.

PR #1017, opened 2023-10-30
ivtest: reformat some regression tests

steveicarus:
This includes:
* always4A
* always4B
* br_gh383a
* br_gh383b
* br_gh383c
* br_gh383d


steveicarus:
> There is loss of information in this patch. Specifically the -pallowsigned=1 from the vlog95 list for br_gh383c and br_gh383d. Also, does the new test support running using strict, forced SV or vlog95? I've not looked at it in enough details to know.

What, really, is the br_gh383c test testing in the regress-vlog95 case? It seems like a CE test with the -pallowsigned=1 flag passed to the compiler. It's easy enough to create a vlog95 test with various options passed to the compiler, but I don't quite know what I'm trying to reproduce here.

PR #1023, opened 2023-11-04
Fix vpi_handle_by_name() for toplevel modules with escaped identifiers

marlonjames:


caryr:
The indenting of the code change is incorrect. Please update it to match what is used for the surrounding code.

Issue #1032, opened 2023-11-15
$left / $right on packed structure field is `x` ?


smunaut:
Reproducer : 

```systemverilog
typedef struct packed {
    logic [7:0] f;
} sbus_t;

module test_top();

    sbus_t sig;

    initial
        $display("%d %d %d", $left(sig.f), $right(sig.f), $bits(sig.f));

endmodule
```

```
$ iverilog -g2012 test.sv -o test && ./test
         x          x          8
```

smunaut:
My attempted fix :

```diff
diff --git a/eval_tree.cc b/eval_tree.cc
index 04e9975e3..79ff2fa0e 100644
--- a/eval_tree.cc
+++ b/eval_tree.cc
@@ -2176,6 +2176,13 @@ static bool get_array_info(const NetExpr*arg, long dim,
        left = param->expr_width() - 1;
        right = 0;
        return false;
+      }
+      if (const NetESelect*select = dynamic_cast<const NetESelect*>(arg)) {
+       const netranges_t&dim_vals = select->net_type()->slice_dimensions();
+       const netrange_t&range = dim_vals[dim-1];
+       left = range.get_msb();
+       right = range.get_lsb();
+       return false;
       }
        /* The argument must be a signal that has enough dimensions. */
       const NetESignal*esig = dynamic_cast<const NetESignal*>(arg);
```

Issue #1033, opened 2023-11-15
Accessing bit values of an enum inside a struct returns enum value instead of bit value


smunaut:
Reproducer :

```systemverilog
typedef enum bit [2:0] {
    ENUM_VAL = 3'b110
} enumtype;

typedef struct packed {
    enumtype t;
} sbus_t;

module test_top();

    sbus_t sig;

    initial
    begin
        sig.t = ENUM_VAL;
        $display("%d %d %d", sig.t[2], sig.t[1], sig.t[0]);
    end

endmodule
```

```
$ iverilog -g2012 ../test.sv -o test && ./test
6 6 6
```

Running in `xsim` shows `1 1 0` which is what I'd expect.

smunaut:
I'm pretty sure the fix will start something like

```diff
diff --git a/elab_expr.cc b/elab_expr.cc
index 1e8e2e502..0c0dacbfd 100644
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -2495,6 +2495,10 @@ static NetExpr* check_for_struct_members(const LineInfo*li,
                  }
                  struct_type = 0;
 
+                 if (!member_comp.index.empty()) {
+                       // Do something clever here ?
+                 }
+
            } else if (const netvector_t*mem_vec = dynamic_cast<const netvector_t*>(member_type)) {
 
                  if (debug_elaborate) {
```

But not sure exactly _what_ to do there :sweat:

smunaut:
Figured it out(I think) see PR.

PR #1034, opened 2023-11-15
Couple of fixes for struct supports

smunaut:
See referenced issues for the problems this PR addresses

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -2486,14 +2486,32 @@ static NetExpr* check_for_struct_members(const LineInfo*li,
 
 	    } else if (const netenum_t*tmp_enum = dynamic_cast<const netenum_t*> (member_type)) {
 
-		    // If the element is an enum, then we don't have
-		    // anything special to do.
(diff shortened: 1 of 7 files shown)


caryr:
Thanks for the patches. Can you please update the eval_tree patch to have the correct spacing (a second level block should be an 8 space tab plus 4 spaces). I do see that the code before the block you added also has a spacing issue and you just followed that, but it would be best to fix the spacing of both your block and the previous block.

We also need to have tests to verify the two fixes are working correctly. It is best to create self checking tests. In general we reserve gold files for tests that we must verify the actual output produced. New tests should be added to the Python based test script.

smunaut:
I wrote the test `.v` files, but I'm not sure in which `.list` file I need to put them TBH.

caryr:
The following covers development topics: https://steveicarus.github.io/iverilog/developer/index.html

The regression section explains how to add tests.

caryr:
This PR resolves #1032 and #1033

PR #1055, opened 2023-12-27
Handle invalid vector slice base expressions

### LINKED ISSUE #1036: [Bug report] Segmentation fault when using out-of-scope variable in slices
Hi there!

I encountered a segmentation fault when using an out-of-scope variable with some slicing.

Here is an example:

`tb.sv`:
```
module testbench;

  bit [31:0] first_words [64 / 32];
  bit [31:0] second_words [32 / 32];

  bit [63:0] first;
  bit [31:0] second;

  top dut(first, second);

  always_comb begin
    for (int first_word_id = 0; first_word_id < 64 / 32; first_word_id++) begin : reset_generator
      first_words[first_word_id] = first[32*first_word_id +: 32];
    end
    for (int second_word_id = 0; second_word_id < 32 / 32; second_word_id++) begin : reset_generator
      second_words[second_word_id] = second[32*first_word_id +: 32]; // Note the usage of first_word_id here
    end
  end

endmodule
```

`top.sv`:
```
module top(first, second);
  output [63:0] first;
  wire [63:0] first;
  output [31:0] second;
  wire [31:0] second;
endmodule
```

Command: `iverilog -g2012 -o out top.sv tb.sv`

Result:
```
tb.sv:16: error: Unable to bind wire/reg/memory `first_word_id' in `testbench.$ivl_for_loop1.reset_generator'
Segmentation fault
```

The bug is still present in the most recent version of Icarus (`c9548f0`), and I'm running on Ubuntu 22.04.

Thanks!
Flavien

larsclausen:
The elaboration of the base expression of a vector slice index can fail and return a nullptr. Currently this results in a nullptr deref. Handle that case by exiting the function early.

Resolves #1036.

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -5994,6 +5994,8 @@ NetExpr* PEIdent::elaborate_expr_net_idx_up_(Design*des, NetScope*scope,
 	    return 0;
 
       NetExpr*base = calculate_up_do_base_(des, scope, need_const);
+      if (!base)
+	    return nullptr;
 
 	// Use the part select width already calculated by test_width().
       unsigned long wid = min_width_;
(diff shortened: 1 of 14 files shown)


steveicarus (approved):
LGTM. The elab_expr() should generate the actual error messages. (The convention is to print the error message as far down the call as possible, and only there, so that an error doesn't lead to a flood of error messages.)

Issue #1052, opened 2023-12-20
Performance issue for a simple generated design with XOR reductions


flaviens:
Hi there!

I noticed that the simulation performance on a simple generated design based on XOR reductions is extremely low in comparison with usual Icarus simulations. Specifically, the speed of `iverilog` is satisfactory, but `vvp` is extremely slow.
The execution of a single evaluation takes more than 2 minutes.

[Here are the sources, with a Readme to reproduce the slow simulation.
](https://github.com/steveicarus/iverilog/files/13728003/slow_xorred.zip)

I'm using the last commit version of Icarus [734f2a0](https://github.com/steveicarus/iverilog/commit/734f2a076bb49bd13ae295cd4e814551338cdbb3).

Thanks!
Flavien

martinwhitaker:
Have you looked at the size of the code you are generating? I'm not surprised it is slow.

larsclausen:
@flaviens what kind of machine are you running this on? When I run your example it takes 2 seconds.

flaviens:
Thank you for your responses!
- @martinwhitaker yes I looked. Simulation is much slower than other random-looking designs of the same size.
- @larsclausen It should be a decent machine (x86, AMD EPYC 7H12).

martinwhitaker:
Running your example on my laptop, it takes 2.4 seconds. But if I add an extra step:
````
 in_data = ~in_data;  
#1;  
````
it takes over 2 minutes.

It's a bit difficult to analyse, because your final `assign` statement is a single line of more than 100,000 characters (which my text editor won't even display!) , but I think the problem is you have created a very large concatenation on both the LHS and RHS. Because it is combinatorial logic, each time one element of the input concatenation is updated, the full output concatenation will be re-evaluated and written to the destination.

I would guess the first step is faster because the initial state is `x`, which will short-circuit some of the intermediate evaluations (in a combinatorial network, only changes propagate to the next node).

If you can split up that final assignment, it should run faster.

larsclausen:
Part of the problem is the structure of how the signals are connected. There are many signals the have the same input signal multiple times, this leads to a exponential number of update events that are being generated. The following example is a more simplified version of this

```SystemVerilog
module test;

  localparam N = 10;
  localparam M = 8;

  bit x[N];
  bit y;

  assign x[0] = y;

  generate for (genvar i = 0; i < N - 1; i++)
    assign x[i+1] = ^{M{x[i]}};
  endgenerate

  initial begin
[...]

PR #1062, opened 2024-01-06
 Use NetReplicate to implement replication for concat

larsclausen:
Currently replication in a concatenation is implemented by simply
concatenating the input signals multiple times by the replication amount.

Replace this to use NetReplicate on the concatenation instead. In case
there is only one input vector to the concatenation the replication will directly
connect to the input vector.

This is slightly more efficient in vvp since the replication functor has
only one input while the concatenation has multiple inputs connected to the
same wire. When an update of the input occurs the replication functor will
only receive a single update, while the concatenation will receive multiple
update events, one for each replication.

### DIFF
--- a/expr_synth.cc
+++ b/expr_synth.cc
@@ -776,23 +776,24 @@ NetNet* NetEConcat::synthesize(Design*des, NetScope*scope, NetExpr*root)
 	    return 0;
       }
 
-	/* Make a NetNet object to carry the output vector. */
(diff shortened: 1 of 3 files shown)


caryr:
Does this fix resolve #1052?

larsclausen:
No, it doesn't. The PR was a result of that discussion, but not the fix. I'll put up a PR with the fix shortly, just need to re-work the failing testcases that rely on a specific order of evaluation.

Issue #1041, opened 2023-12-07
VPI: vpi_put_value() to a WIRE is not sticky.


gatk555:
By sticky, I mean the behaviour described in the standard (I only have 2001, section 27.32 describes the function):

`When vpi_put_value() is called for an object of type vpiNet or vpiNetBit, and with modes of vpiInertialDelay, vpiTransportDelay, vpiPureTransportDelay, or vpiNoDelay, the value supplied overrides the resolved value of the net. This value shall remain in effect until one of the drivers of the net changes value. When this occurs, the net shall be re-evaluated using the normal resolution algorithm
`

That does not happen.  A minimal test case (attached) has an undriven wire whose value is copied to a register on a clock edge. The wire is set to 1 from VPI at the start of simulation, so my interpretation of the above is that the register value should become 1. It is z.

If vpiNoDelay is replaced by vpiForceFlag it works as expected. But that is no good if there are drivers on the wire. In my application it may be the avatar of an inout port of the top module ([described here](https://github.com/steveicarus/iverilog/discussions/1035)).

I tried to find out why this happens, but ran aground around line 1077 in net_vvp.h.  Too much C++++++ for me!

[test_case.zip](https://github.com/steveicarus/iverilog/files/13606019/test_case.zip)

martinwhitaker:
I agree, this does appear to be incorrect behaviour.

martinwhitaker:
This happens because the start of simulation callbacks are being executed before the simulator performs its initial constant propagation. An undriven wire is connected to a constant Z, and that Z value gets propagated after the callbacks, replacing the value set by vpi_put_value(). I have a fix for this, but it is causing 3 failures in the test suite which I need to investigate.

gatk555:
Thank you for looking at it.  Today, I tried digging a bit deeper and came to much the same conclusion, although I have not understood what is happening in schedule.cc.  What I have found is that changing cbStartOfSimulation to cbNextSimTime fixes the problem in my demonstration code. I shall try that in my application code.

After looking at the documents I have, I am no longer convinced this is a bug.  The timing of these callbacks relative to both initial blocks and implicit initialisations seems undefined. Perhaps later specifications are better at that.

martinwhitaker:
The SystemVerilog simulation reference algorithm in IEEE 1800-2017 section 4.5 does indicate that all nets and variables are initialised before any events are executed, and in section 4.3 it does state that VPI callbacks are evaluation events.

[...]

PR #1065, opened 2024-01-07
vvp undriven net initialisation

martinwhitaker:
This is my proposed fix for #1041. Putting it up for review in case anyone can think of a better way (i.e. not requiring a vvp API change).


gatk555:
I pulled this branch and merged my local changes. It certainly seems to fix #1041. I still have a problem with VPI inputs apparently failing to propagate through ASSIGN, but as attempts to reduce it to a simple test case failed, it is probably my error.

Thanks, G

gatk555:
It seems that my problem with ASSIGNed nets occurs because this PR has uncovered a latent bug in schedule_simulate().
In the attached example, the value of an undriven wire, w, is passed through a gate (using ASSIGN) to another wire that is monitored.  The wire w is then driven periodically by VPI.  With a build of current master, it works as expected, but #1041 shows up as value 'x' rather than '1' at time zero.  With this PR, there is no change at all in the output wire.

The fix seems simple: move the assignment "sim_started = true;" before the call to vpiStartOfSim(). (Patch file included.)
It seems that without that, the first update of the output wire is stranded on schedule_init_list and blocks all later changes.  The regression scripts vvp_reg.pl and vpi_reg.pl still pass with this change.

[test_case2.zip](https://github.com/steveicarus/iverilog/files/13927419/test_case2.zip)

martinwhitaker:
@gatk555 , agreed, that is a latent bug and your fix looks correct. Do you want to be formally credited as the author? If so, we'll either need a git-formatted patch or a separate PR  (but remove the old line, don't just comment it out). If not, I can just give you credit in the commit message.

gatk555:
The simplest thing would be for you to add it to your PR as an additional change, incidental credit is fine.  I did consider making a PR of it, but either that has to wait until yours is committed, or I would have to use your branch as a base, and that would look like stealing.  As a stand-alone change, it does nothing useful, as far as I can tell.

Thanks,
G.

larsclausen:
@martinwhitaker  You can `git commit --author="..."` to assign authorship to somebody else.

martinwhitaker:
@gatk555 , actually it is useful standalone. If you change wire w in your test case to a reg, you'll see the same bug on the master branch. But I'll add it to this PR. Let me know if you want me to use anything other than your GitHub username.

gatk555:
The username is fine.
Thanks, G.

steveicarus (approved):

PR #1066, opened 2024-01-07
vvp: Improve concat performance

### LINKED ISSUE #1052: Performance issue for a simple generated design with XOR reductions

larsclausen:
A concat typically has multiple inputs. Whenever one of the input values
change the output value of the concat is updated and propagated to its
downstream consumers.

When multiple inputs change within the same cycle each input will cause a
update propagation. Depending of the overall structure of the design this
can cause a significant performance penalty.

E.g. the following synthetic structure has a exponential runtime increase
based on the value of N.

```SystemVerilog
reg [N-1:0] x;
generate for (genvar i = 0; i < N - 1; i++)
  assign x[i+1] = ^{x[i],x[i]};
endgenerate
```

To improve this defer the value propagation of the concat to the end of the
current cycle, this allows multiple input updates to be included in a
single output update.

For the example in report #1052 this reduced the runtime from 2 minutes to
essentially 0.

Also replace individual bit access when copying the vector with the the set_vec()
function, which can copy data word by word. This results in a small linear
additional performance improvement.

There is a small number of test cases that rely on the non-deferred updated
of the concat operation and fail when switching to deferred updates. These
test cases are updated to be race condition free.

Resolves #1052.

### DIFF
--- a/ivtest/ivltests/br_gh497a.v
+++ b/ivtest/ivltests/br_gh497a.v
@@ -25,6 +25,7 @@ assign array6[2:1] = 8'h32;
 reg failed = 0;
 
 initial begin
+  #0
   $display("%h", array1);
   if (array1 !== 16'h4321) failed = 1;
   $display("%h", array2);
--- a/ivtest/ivltests/part_sel_port.v
+++ b/ivtest/ivltests/part_sel_port.v
@@ -33,6 +33,7 @@ mod_test dut(test_string[1:8]);
 mod_test2 dut2(test_string[9:16]);
 
 initial begin
+    #0
     if(test_string !== "testTESTabcdefgh") begin
         $display("FAILED");
         $finish();
--- a/ivtest/ivltests/pr1002.v
+++ b/ivtest/ivltests/pr1002.v
@@ -8,29 +8,11 @@ assign dataout = datain >>> 2;
 
 reg test_failed;
 
-initial
-  begin
-     test_failed = 0;
-     #1 datain = 14'h0FFF;
-     #1 datain = 14'h0000;
-     #1 datain = 14'h1FFF;
-     #1 datain = 14'h1000;
-     #1 datain = 14'h2FFF;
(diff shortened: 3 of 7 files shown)


steveicarus:
Is it really true that values assigned to nets propagate in delta time, instead of instantly? For example, is this:
```
wire foo = bar;
reg bar;
initial begin
  bar = 1;
  $display(foo, bar);
end
```
a race?
And if a "#0" is in front of $display, is it still a race?

larsclausen:
It for sure is a race condition if there is no #0. That's pretty much the example in section "4.8 Race conditions" in the LRM.
If there is a #0 I believe it is not supposed to be a race condition.

steveicarus:
Does #0 instead of #1 in the tests you added still run correctly? If so, that would be a tighter test, I think.

ivtest/ivltests/br_gh497a.v
```diff
@@ -25,6 +25,7 @@ assign array6[2:1] = 8'h32;
 reg failed = 0;
 
 initial begin
+  #1
```
steveicarus, line 28: Can this be a #0 instead of #1?

steveicarus (approved):
LGTM, reference some minor comments.

larsclausen:
changed all the delays to `#0`.

PR #1068, opened 2024-01-13
Configure option --enable-libvvp allows vvp to be used a shared library

gatk555:
This is essentially the same as #667, with the same motivation: a VVP simulation may be embedded in another program. It avoids code duplication and adds some documentation, as requested for #667.


steveicarus:
This looks promising.

Do we need to add something to the regression tests to verify that it continues to work properly?

gatk555:
During development I used modified versions of the test.yml script to check operation on Windows and Mac. As this change is (currently, but see my [comment](https://github.com/steveicarus/iverilog/pull/667#issuecomment-1841654384)  on #667) enabled in configure, the only good way to regression test seems to be to double-up test.yml, either by adding extra jobs or a parallel script. A second script seems the best to me.  I am happy to add that to the PR.

Slightly off-topic but may be interesting: the enclosed file shows results of a SPICE simulation of an ADC circuit. The digital part of the ADC was an instance of libvvp.so.

[adc.svg.zip](https://github.com/steveicarus/iverilog/files/13997584/adc.svg.zip)

steveicarus:
@larsclausen, @martinwhitaker , @caryr
Can you all comment on this PR? I'm inclined to accept it, but it does add some build-time complexity to a build that is already a bit complex. Also, testing remains an open issue.

steveicarus (approved):
Thank you for the documentation. That makes this easier to accept. I believe we may want to think harder about how to test this, and that conversation, and whether this is the best way to deploy the feature, may lead to further issues created.

martinwhitaker:
This looks good to me. Having struggled to make a ngspice/iverilog co-simulation work in the past, something that makes it easier gets my vote!

For testing, how about after the current tests, reconfigure/rebuild with --enable-libvvp then rerun just the vpi tests. That should give reasonable coverage of the libvvp API. I doubt we fully cover the vvp command line options in CI as it is.

martinwhitaker (approved):

larsclausen (approved):

vvp/libvvp.h
```diff
@@ -0,0 +1,76 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* The first three functions may be called at any time.
+ * vvp_set_stop_is_finish(true) is equivalent to vvp's "-n" option.
+ */
+
+extern void vvp_set_stop_is_finish(bool flag);
```
larsclausen, line 18: might want to add a include to stdbool.h

caryr:
I agree this would be a nice addition and while I don't think a test is required immediately can we restructure the existing vvp to use the dynamic library so it tests this by default?

martinwhitaker:
> I agree this would be a nice addition and while I don't think a test is required immediately can we restructure the existing vvp to use the dynamic library so it tests this by default?

IIUC, that is done when you build with the --enable-libvvp option, hence my suggestion for testing it. I don't think we need to rerun the full set of tests, just enough to prove the library interface is working. I suggested the vpi tests because that covers the functions for loading vpi modules.

caryr:
I should have read the documentation changes more closely.

Now that I have done that one thing I really do not like is the restriction of a single running instance. I believe all the static variables need to be moved into the stub or handled correctly per process.

One of my goals for the 14 release is to support running our regression tests in parallel based on the machines actual core count to reduce the total test time and in the past I usually ran multiple circuit regressions at the same time so switching to a single running instance requirement would be a no go from my side. I believe even with the SPICE use case people will eventually run multiple instances and we don't want things crashing or behaving oddly when this happens.

I also do not like how vvp/main.cc was moved and then copied back since it has now lost all the history that used to be associated with the file. I get the intent was that lib_main was the new replacement so the logs should go with that, but that's not how it ended up.

caryr (changes requested):
See my comment in the main thread, but the short list is I do not think requiring a single running instance on a system is acceptable (this is also across users since a .so is a global item) and I do not like that vvp/main.cc no longer has the revision history.

martinwhitaker:
> See my comment above, but the short list is I do not think requiring a single running instance on a system is acceptable (this is also across users since a .so is a global item)

Are you sure about this? To the best of my knowledge, each process gets its own copy of the static data associated with a dynamic library. Anything else would be a massive security hole.

I think what the documentation is saying is that you can't have two separate simulations running in the same process. That's an acceptable limitation to me.

> and I do not like that vvp/main.cc no longer has the revision history.

I checked out @gatk555's libvvp branch. `git log vvp/main.cc` shows history going back to 2001. `git log --follow vvp/lib_main.cc` shows the same history and `git blame vvp/lib_main.cc` shows what you would hope for. So the only problem seems to be that `git blame vvp/main.cc` doesn't show what you would expect.

caryr:
I certainly could have misread and I agree limiting to a single instance per process is fine.

gatk555:
The blame problem is fixable.  After some testing I shall push a replacement version.  I also added a suggestion to use --follow to lib_main.cc's log entry and it should remove the merge conflict.

steveicarus:
After reading @caryr 's comments carefully, and looking at the posted documentation, I see the problem. If a single program were to try to run a simulation twice, the second run, even if sequential, would carry leftover memory values from the first run. This strikes me as very messy, and a great source of mysterious errors. Can we at least fix up the library to assert that doesn't happen?

martinwhitaker:
In the longer term, vvp already has code to release all allocated memory (conditionally compiled with `CHECK_WITH_VALGRIND`). So it might not be too much work to support sequential runs in the same process if that is seen as a desirable feature. We would need to run with valgrind occasionally to properly test it, which is another thing not currently supported by the new test framework.

caryr:
> In the longer term, vvp already has code to release all allocated memory (conditionally compiled with `CHECK_WITH_VALGRIND`). So it might not be too much work to support sequential runs in the same process if that is seen as a desirable feature. We would need to run with valgrind occasionally to properly test it, which is another thing not currently supported by the new test framework.

It tries to release all the code, but there are cases where a few things get missed. When I added the valgrind support it was mostly to make sure we did not have any place where we used undefined values and to make sure there were no memory leaks. The later required releasing as many of the items as we can to get the exception list to a manageable size. I still run valgrind from time to time, but yes, it uses the old infrastructure. I hope to update the new Python code to support all the functionality in the original script(s), but that may require some refactoring and I'm still behind on things I committed to fix before the next release.

caryr:
Are we thinking this is something we should just merge for the 14 release even though it is not tested as part of the normal flow so the changes are available for use in a mixed signal simulation? I just want to make sure we do not pick up additional support work because we allowed code that was not fully vetted into the repo.

caryr (approved):

gatk555:
Thanks!

PR #1069, opened 2024-01-14
ivtest: Remove outdated pr1963962 SystemVerilog mode gold file

larsclausen:
Starting with commit 96df251c955a ("Suppress unnecessary VCD/LXT/LXT2 warnings about packages.") there is no longer a warning printed that the unit scope can't be printed if it is empty.

Remove the special SystemVerilog mode gold file for the pr1963962 test that expects this warning.


martinwhitaker:
Oops, missed that. I miss having the expected results files to diff against...

I can't see why the CI has failed on the VPI tests.

larsclausen:
It looks like the VPI test failure on Mac is unrelated. I guess something changed in the environment.

PR #1072, opened 2024-01-15
iverilog-vpi.sh: Put `IVCXX` in quotes to allow to pass arguments

larsclausen:
`IVCXX` which contains the C++ compiler that will be invoked when building an vpi module might contain additional arguments that get passed to the compiler. E.g. as the C++ version (`-std=c++11`). For this to work properly IVCXX needs to be put in quotes.

This fixes intermittent CI failures for the MacOS target.


martinwhitaker:
Looks good to me.

Issue #483, opened 2021-02-09
$bits on item in a generate block fails when used in a constant expression


zachjs:
```SystemVerilog
module top;
	parameter ENABLE = 1;
	if (ENABLE) begin : blk
		wire [7:0] w;
	end
	wire [7:0] x;
	wire [$bits(blk.w)-1:0] y;
	wire [$bits(x)-1:0] z;
	initial begin
		$display("blk.w: %b (%0d bits)", blk.w, $bits(blk.w));
		$display("x: %b (%0d bits)", x, $bits(x));
		$display("y: %b (%0d bits)", y, $bits(y));
		$display("z: %b (%0d bits)", z, $bits(z));
	end
endmodule
```

When simulating the above test case, most tools produce:
```
blk.w: zzzzzzzz (8 bits)
x: zzzzzzzz (8 bits)
y: zzzzzzzz (8 bits)
z: zzzzzzzz (8 bits)
```

However, iverilog (both version 11.0 and the latest master) produces:
```
blk.w: zzzzzzzz (8 bits)
x: zzzzzzzz (8 bits)
y: xx (0 bits)
z: zzzzzzzz (8 bits)
```
Notice that `$bits(blk.w)` correctly evaluates to `8` in the `$display` call, but not in the declaration of `y`.

caryr:
I can confirm this is failing with development.

caryr:
This are a couple issues around this bug. First generate schemes appear to be elaborated after the enclosing scope. There is also no error reported if the $bits() function is given an invalid/non-existed identifier, the elaboration order is also dependent on the actual name so this can fail even when no generate scope is used. I need to figure out how to trigger an elaboration of the dependent signals to resolve this.

zachjs:
I've encountered a distinct but probably related issue on the below case. I'd be happy to file this separately if you'd prefer.
```SystemVerilog
module Module;
    parameter T = 10;
    wire [T-1:0] x;
    initial $display("Module %b %0d", x, T);
endmodule

module top;
    wire [7:0] x;
    Module #($bits(x)) mA();
    Module #(8) mB();
endmodule
```

Most tools yield
```
Module zzzzzzzz 8
Module zzzzzzzz 8
```
whereas iverlog yields
```
Module xx 0
Module zzzzzzzz 8
```

caryr:
Yes, that looks like the same issue. I will make sure whatever fix I come up with fixes both.

caryr:
[...]

PR #1102, opened 2024-02-19
Require declaration before use for parameters, nets, variables, and named events

martinwhitaker:
Traditionally, when not using hierarchical names, Verilog has required declaration before use for parameters, nets, variables, and named events, apart from the special case of single bit implicit nets . This was never explicitly stated in any of the IEEE 1364 revisions, but was in early revisions of IEEE 1800 (although that text has been lost in later revisions), and most other simulators require it.

Icarus Verilog has always allowed use before declaration. Mostly this is relatively harmless, allowing it to accept code that other tools will reject. It is possible to contrive a test case that will result in different behaviour, e.g.
````
localparam w = 8;

task t;
  reg [w:1] v;
  localparam w = 2;
endtask;
````
where `iverilog` will make `v` 2 bits and other tools will make it 8 bits.

There is a bug report for this on the old SourceForge bug tracker, but as it is a change to long-standing behaviour, I'd like to get all the developers' approval for it.

Note that I had to change 7 tests in the test suite for this:
* sdf_interconnect[1-4] just needed the code reordering.
* pr1909940 and pr1909940b are designed to test use before implicit declaration, so I changed them to CE.
* for vhdl_fa4_test4 the VHDL pre-processor generates code that uses parameters before they are declared, so I changed it to NI.

### DIFF
--- a/PEvent.cc
+++ b/PEvent.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2004-2019 Stephen Williams (steve@icarus.com)
+ * Copyright (c) 2004-2024 Stephen Williams (steve@icarus.com)
  *
  *    This source code is free software; you can redistribute it
  *    and/or modify it in source code form under the terms of the GNU
@@ -21,8 +21,8 @@
 
 # include  "PEvent.h"
 
-PEvent::PEvent(perm_string n)
-: name_(n)
+PEvent::PEvent(perm_string n, unsigned lexical_pos)
+: name_(n), lexical_pos_(lexical_pos)
 {
 }
 
--- a/PEvent.h
+++ b/PEvent.h
@@ -1,7 +1,7 @@
 #ifndef IVL_PEvent_H
 #define IVL_PEvent_H
 /*
- * Copyright (c) 2000-2019 Stephen Williams (steve@icarus.com)
+ * Copyright (c) 2000-2024 Stephen Williams (steve@icarus.com)
  *
  *    This source code is free software; you can redistribute it
  *    and/or modify it in source code form under the terms of the GNU
@@ -36,10 +36,12 @@ class PEvent : public PNamedItem {
     public:
 	// The name is a perm-allocated string. It is the simple name
 	// of the event, without any scope.
-      explicit PEvent(perm_string name);
+      explicit PEvent(perm_string name, unsigned lexical_pos);
       ~PEvent();
 
       perm_string name() const;
+
+      unsigned lexical_pos() const { return lexical_pos_; }
 
       void elaborate_scope(Design*des, NetScope*scope) const;
 
@@ -47,6 +49,7 @@ class PEvent : public PNamedItem {
 
     private:
       perm_string name_;
+      unsigned lexical_pos_;
 
     private: // not implemented
       PEvent(const PEvent&);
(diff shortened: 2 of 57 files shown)


larsclausen (approved):
Looks good. 
Small typo in the last commit message "rgression"

caryr (approved):
Is it possible to fix the VHDL compiler and more importantly can we generate a proper use before declaration message to help people deal with the change? To me it seems like not declared/found should only be used for variables that are not declared or found in the scope.

I'll approve for now, but I strongly feel an order mismatch message should be reported instead of the current message to make it easy for our users to understand what is going on..

Also, does it properly understand things on the same line for example `logic foo; assign foo=1;` is okay, but `assign foo=1; logic foo;` is not? No user should write code like this, but an automatic code generator could.

martinwhitaker:
> Is it possible to fix the VHDL compiler

I've never looked at the VHDL pre-processor, so don't know how hard that would be. The problem is that it is using named constants in a module's port declarations, but is declaring the constants as localparams inside the module.

> and more importantly can we generate a proper use before declaration message to help people deal with the change? To me it seems like not declared/found should only be used for variables that are not declared or found in the scope.

I was planning on doing that, but wanted to check that this PR was acceptable in principle before doing any more work. I will also look at genvars at some point, but my primary reason for doing this work is to enable me to finish my fix for issue #483 et al.

> Also, does it properly understand things on the same line for example `logic foo; assign foo=1;` is okay, but `assign foo=1; logic foo;` is not? No user should write code like this, but an automatic code generator could.

Yes, that's why I added the `lexical_pos` field to the `vlltype` struct, rather than simply using the line number. The `lexical_pos` is incremented each time the scanner returns an identifier token. We can increment `lexical_pos` more often without breaking anything if we subsequently find a need to know the lexical order of other items (although we might need to make it a `uint64_t` if we go to extremes.

caryr:
> > Is it possible to fix the VHDL compiler
> 
> I've never looked at the VHDL pre-processor, so don't know how hard that would be. The problem is that it is using named constants in a module's port declarations, but is declaring the constants as localparams inside the module.

Sounds like they should be using a parameter port list for the ANSI port list.
 
> > and more importantly can we generate a proper use before declaration message to help people deal with the change? To me it seems like not declared/found should only be used for variables that are not declared or found in the scope.
> 
> I was planning on doing that, but wanted to check that this PR was acceptable in principle before doing any more work. I will also look at genvars at some point, but my primary reason for doing this work is to enable me to finish my fix for issue #483 et al.

Sounds good.

> > Also, does it properly understand things on the same line for example `logic foo; assign foo=1;` is okay, but `assign foo=1; logic foo;` is not? No user should write code like this, but an automatic code generator could.
> 
> Yes, that's why I added the `lexical_pos` field to the `vlltype` struct, rather than simply using the line number. The `lexical_pos` is incremented each time the scanner returns an identifier token. We can increment `lexical_pos` more often without breaking anything if we subsequently find a need to know the lexical order of other items (although we might need to make it a `uint64_t` if we go to extremes.

Per token should be good enough and if not then we could switch to per character and then reset at the start of each line. Do you keep the token count for include files which could hide declarations or usage. I will confess I didn't look at the code in 100% detail, because I trust you coded it correctly. I just skimmed to see that it made sense and am thinking about corner cases that could break things.

martinwhitaker:
@steveicarus , I'll assume you have no objection and merge this soon.

PR #1115, opened 2024-04-12
Update handling of single-element arrays.

### LINKED ISSUE #1113: vvp: array.cc:1561: void compile_array_alias(char*, char*, char*): Assertion `mem' failed.
The following:

```
module __sample__main(
  input wire [48:0] x1,
  output wire [48:0] out
);
  wire [48:0] x17[0:0];
  assign x17[0] = x1;
  assign out = {x17[0]};
endmodule
```

triggers the following vvp error for me at ef7f0a8f38782dfc0872b1e352ccf32343c10bb8:

```
vvp: array.cc:1561: void compile_array_alias(char*, char*, char*): Assertion `mem' failed.
```

grebe:
Also, add a test. This fixes #1113.

### DIFF
--- a/ivtest/ivltests/single_element_array.v
+++ b/ivtest/ivltests/single_element_array.v
@@ -1,0 +1,31 @@
+module SingleElementArray(
+  input wire [48:0] x1,
+  output wire [48:0] out
+);
+  wire [48:0] x17[0:0];
+  assign x17[0] = x1;
+  assign out = {x17[0]};
+endmodule
+
+module testbench;
+  reg [48:0] in;
+  wire [48:0] out;
+
+  SingleElementArray dut(.x1(in), .out(out));
+
+  initial begin
+    in = 49'h0000000000000;
+    #1;
+    if (out != 49'h0000000000000) begin
+      $display("FAILED");
+      $finish;
+    end
+    in = 49'h1555555555555;
+    #1;
+    if (out != 49'h1555555555555) begin
+      $display("FAILED");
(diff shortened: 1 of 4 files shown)


grebe:
I'm new to the codebase and only sort-of understand what this is doing, so feel free to give feedback if I should do something else.

martinwhitaker:
From vvp's point of view your fix converts the array into a single vector. This gives the correct behaviour for simulation, but does affect the VPI view of the design. I think a better fix would be
````
--- a/tgt-vvp/vvp_scope.c
+++ b/tgt-vvp/vvp_scope.c
@@ -697,7 +697,8 @@ static void draw_net_in_scope(ivl_signal_t sig)
                       so the word count for the signal and the alias
                       *must* match. */
 
-                 if (word_count == ivl_signal_array_count(nex_data->net)) {
+                 if ((ivl_signal_dimensions(nex_data->net) > 0) &&
+                      word_count == ivl_signal_array_count(nex_data->net)) {
                    if (iword == 0) {
                      fprintf(vvp_out, "v%p .array \"%s\", v%p; Alias to %s \n",
                              sig, vvp_mangle_name(ivl_signal_basename(sig)),
````
Note that the `if (iword == 0)` must be retained, otherwise the alias will be repeated for every word in the array. It's hard to demonstrate that though, as AFAICT the compiler never generates an alias for an entire array.

martinwhitaker:
Merged. Thanks for your contribution.

PR #1190, opened 2024-12-21
iverilog: Allow tool path to have spaces in it

### LINKED ISSUE #1187: Iverilog doesn't accept spaces in the value path of the -B flag.
Iverilog exists with an error if the value  of ``-B`` has a space in the path,  but do accept correctly space in source files paths.  Is there a workaround (escaping?) for passing to ``-B`` a value with a space?

I am using  ``Icarus Verilog version 13.0 (devel) (s20221226-533-g676b36e45`` on Mac OSX apple silicon.

```
/Users/user/.apio/packages/tools-oss-cad-suite/bin/iverilog -B '/Users/user/xx packages/tools-oss-cad-suite/lib/ivl' -o _build/ledon_tb.out -DVCD_OUTPUT=_build/ledon_tb -DNO_ICE40_DEFAULT_ASSIGNMENTS /Users/user/xx_packages/tools-oss-cad-suite/share/yosys/ice40/cells_sim.v main.v ledon_tb.v
sh: /Users/user/xx: No such file or directory
sh: /Users/user/xx: No such file or directory
```

larsclausen:
When invoking the various subcommands make sure the path to the executable is quoted. This allows it to contain spaces.

On Windows quoting the command passed to `system()` or `popen()` is not enough.  The default behavior under windows will strip quotes from the beginning and end of the command string under certain conditions. For this reason implement `system()` and `popen()` equivalents that do not suffer from this.

Resolves #1187

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -59,6 +59,8 @@ SUBDIRS += driver-vpi
 # rebuild the lexor_keyword.cc file. If we do, then we want to use the
 # local version instead of the one is $(srcdir).
 vpath lexor_keyword.cc .
+vpath %.c $(srcdir)/libmisc
+vpath %.c $(srcdir)
 vpath %.cc $(srcdir)/libmisc
 vpath %.cc $(srcdir)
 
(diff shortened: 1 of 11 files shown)


caryr:
Can we add the basics of this to quote the paths so we can resolve the base issue for some of the OSes?

caryr:
Maybe try using vpath in the Makefile like in the main and vvp directories. That may also make cppcheck work without issue.

PR #1212, opened 2025-02-19
Using C++11 threads instead of pthread

gian21391:
I have ported vvp (and related shared libraries) to MSVC. To do that, I decided to remove the dependency from pthread and use the threading support offered by C++11.

I think it makes sense to have this also in the main repository.

### DIFF
--- a/vpi/vcd_priv2.cc
+++ b/vpi/vcd_priv2.cc
@@ -21,11 +21,13 @@
 # include  <map>
 # include  <set>
 # include  <string>
-# include  <pthread.h>
 # include  <cstdlib>
 # include  <cstdint>
 # include  <cstring>
 # include  <cassert>
+# include  <thread>
+# include  <mutex>
+# include  <condition_variable>
 
 /*
    Nexus Id cache
@@ -84,7 +86,7 @@ extern "C" void vcd_scope_names_delete(void)
       vcd_scope_names_set.clear();
 }
 
-static pthread_t work_thread;
+static std::thread work_thread;
 
 static const unsigned WORK_QUEUE_SIZE = 128*1024;
 static const unsigned WORK_QUEUE_BATCH_MIN = 4*1024;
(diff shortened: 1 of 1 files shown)


vpi/vcd_priv2.cc
```diff
@@ -223,10 +223,8 @@ extern "C" void vcd_work_sync(void)
 	    end_batch();
 
       if (work_queue_fill > 0) {
-	    pthread_mutex_lock(&work_queue_mutex);
-	    while (work_queue_fill > 0)
-		  pthread_cond_wait(&work_queue_is_empty_sig, &work_queue_mutex);
-	    pthread_mutex_unlock(&work_queue_mutex);
+            std::unique_lock<std::mutex> lock(work_queue_mutex);
+            work_queue_is_empty_sig.wait(lock, []{ return !(work_queue_fill > 0); });
```
larsclausen, line 227: maybe 
```suggestion
            work_queue_is_empty_sig.wait(lock, []{ return work_queue_fill == 0; });
```

larsclausen (approved):
This looks great, thanks!

caryr:
I would need to check, but I think the files we take directly from GTKWave (specifically fstapi.c) currently use pthreads so to get this completely clean we would have to get Tony to change this to C++ on his end as well. If we provide a patch that may be enough. I'm not against using C++ threads for the C++ code, but we need to remember there is also C code that can only use pthreads so do we really save anything by switching the C++ code?

gian21391:
> I would need to check, but I think the files we take directly from GTKWave (specifically fstapi.c) currently use pthreads so to get this completely clean we would have to get Tony to change this to C++ on his end as well. If we provide a patch that may be enough. I'm not against using C++ threads for the C++ code, but we need to remember there is also C code that can only use pthreads so do we really save anything by switching the C++ code?

Here the point is to have a working version of vvp that I can compile using MSVC, rather than removing the dependency per se. That works perfectly without enabling the parallel writer.

I have now also fixed the bug related to the LXT2 writer: the LXT2 writer invoked `vcd_work_terminate()` in 2 separate places, which is invalid when using std::thread since all resources are freed when joining.

caryr:
Do you know if the performance is similar? Either anecdotal reports or (preferred) actual test results.

Also is the finish_cb() always called so we know vcd_work_terminate() never needs to be called from close_dumpfile()?

PR #1259, opened 2025-07-06
Fix issues with waveform viewer documentation and mention Surfer

oscargus:
There were some inconsistencies in the documentation regarding file formats. "Use the "-fst" extended argument to
activate LXT output" etc.

I also took the liberty to mention our waveform viewer Surfer and therefore renamed the page to "Viewing Waveforms". Still, examples consider GTKWave and there is no Surfer screen shot etc. Still, as Surfer runs on a few platforms where it is hard to get GTKWave to run, I thought it was useful for Icarus Verilog users to know that there are options.


caryr:
I want to install and try surfer before approving this change.

PR #1262, opened 2025-07-12
Fix regression tests to not $error-fail assertions

wsnyder:
This fixes several `assert`/`assume` tests to not $error-fail, as this causes other simulators to fail https://github.com/chipsalliance/sv-tests

If you prefer to do this differently, of feel free of course.  Alternatively let me know if prefer the tests stay as-is and I'll get them blacklisted on sv-tests.

Thanks

### DIFF
--- a/ivtest/ivltests/sv_immediate_assert.v
+++ b/ivtest/ivltests/sv_immediate_assert.v
@@ -4,7 +4,7 @@ integer i = 1;
 
 initial begin
   assert(i == 1);
-  assert(i == 0);
+  // Fail: assert(i == 0);
   assert(i == 1) else $display("Check 3 : this shouldn't be displayed");
   assert(i == 0) else $display("Check 4 : this should be displayed");
   assert(i == 1) $display("Check 5 : this should be displayed");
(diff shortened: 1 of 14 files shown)


ivtest/ivltests/sv_immediate_assert.v
```diff
@@ -4,7 +4,7 @@ integer i = 1;
 
 initial begin
   assert(i == 1);
-  assert(i == 0);
```
martinwhitaker, line 7: This reduces our fault coverage. Nothing else tests that an assert with no pass or fail statement behaves correctly.
Ditto for the other tests.

ivtest/ivltests/sv_immediate_assert.v
```diff
@@ -15,8 +15,8 @@ initial begin
     else $display("Check 8 : this should be displayed");
 
   a_i_is_non_0 : assert(i == 0)
-    $display("Check 9 : this shouldn't be displayed");
-    else $error("Check 9 : this should be displayed");
```
martinwhitaker, line 19: My reason for using $error here was to make it clear that the message came from the fail statement.

martinwhitaker (commented):
Let's see what @caryr has to say, but I'd prefer not to make these changes.

wsnyder:
No problem, I'll handle outside ivtests.

caryr:
I've commented in the sv-tests repo pull request. To fully understand this I would need more context regarding why you wanted to change this. For example was it to try and make the gold files match because all simulators should fail when a `$error()` is executed, though they should not fail immediately like `$fatal()`.

caryr:
I'll take a look to see how things are working.

Issue #917, opened 2023-05-06
Floating point formatting test failures on aarch64


jleightcap:
On both the 11.0 and 12.0 release, two tests fail on `aarch64` due to floating point formatting:

```
                    br1029a: ==> Failed - output does not match gold file.
                    br1029b: ==> Failed - output does not match gold file.
```

- https://github.com/steveicarus/ivtest/blob/master/ivltests/br1029a.v
- https://github.com/steveicarus/ivtest/blob/master/ivltests/br1029b.v

Using the current 11.0 macos release installed through brew,
```
$ iverilog -v
Icarus Verilog version 11.0 (stable) ()
# ...
$ which iverilog
/opt/homebrew/bin/iverilog
$ iverilog source/ivtest/ivltests/br1029a.v
$ ./a.out
a: 0.4  0  0  0
b: 0.5  1  1  1
c: 0.6  1  1  1
d: 2.4  2  2  10
e: 2.5  3  3  11
f: 2.6  3  3  11
a: -0.4  -0  0  0
b: -0.5  -1  0  0
c: -0.6  -1  0  0
d: -2.4  -2  0  0
e: -2.5  -3  0  0
f: -2.6  -3  0  0
```

And using a fresh build of the 12.0 release through nix,

```sh
$ ./result/bin/iverilog source/ivtest/ivltests/br1029a.v
$ ./a.out
a: 0.4  0  0  0
b: 0.5  1  1  1
c: 0.6  1  1  1
d: 2.4  2  2  10
e: 2.5  3  3  11
f: 2.6  3  3  11
a: -0.4  -0  0  0
b: -0.5  -1  0  0
c: -0.6  -1  0  0
d: -2.4  -2  0  0
e: -2.5  -3  0  0
f: -2.6  -3  0  0
```

In both cases the formatting of `%0x` and  `%0b` differs from the spec file.

See our [nixpkgs discussion](https://github.com/NixOS/nixpkgs/pull/229364).

Thanks, let me know if I'm missing something and I'm happy to provide more details as needed :slightly_smiling_face:

martinwhitaker:
This looks to be a architecture-dependent bug in the vpi_get_value() implementation when called with a vpiBinStrVal or vpiHexStrVal format. But I don't have an aarch64 machine, so can't investigate further.

larsclausen:
The code looks like this `snprintf(rbuf, RBUF_SIZE, "%" PRIx64, (uint64_t)vlg_round(rval));`

[...]

PR #1264, opened 2025-07-29
Fix and refactor conversion from `double` to `uint64_t` (fixes test br1029b on x86-64-v4)

weiren2:
## Issue
When compiling with `-march=x86-64-v4` or `-march=znver4`, test case `br1029b` fail to pass. Based on PR #919 and discussion in #917, I think the following https://github.com/steveicarus/iverilog/blob/db82380cecf9943fcc397818e6899b7146442127/vvp/vthread.cc#L2383-L2386 is the root cause. Basically there are still direct double-to-unsigned-integer casts (undefined behavior) in `vpi_vthr_vector.cc` and `vthread.cc`.

## This PR fixes:
- Replaced custom rounding with rounding functions from standard library (`<cmath>` is already included in the code and I noticed that c++11 is in the default flags).
- Used a common utility header for `vlg_round_to_u64` to safely cast form `double` to `uint64_t` and reuse across files.
- Refactor binary string value conversion logic that used in `vpi_callback.cc` and `vpi_vthr_vector.c`.

I have tested this PR locally on my x86 machine both with and without `-march=x86-64-v4`. Test case `br1029b` passes in both configurations, and no regressions were introduced. It probably needs to be tested on other architectures as well but I don't have ARM machines at the moment.

Please let me know if there are any comments and feel free to make edits.

### DIFF
--- a/vvp/vpi_callback.cc
+++ b/vvp/vpi_callback.cc
@@ -987,28 +987,26 @@ static void real_signal_value(struct t_vpi_value*vp, double rval)
 	    break;
 
 	  case vpiBinStrVal: {
-		uint64_t val = vlg_round_to_u64(rval);
-		unsigned len = 0;
+        const uint64_t val = vlg_round_to_u64(rval);
+        unsigned len = 0;
 
-		while (val > 0) {
-		      len += 1;
-		      val /= 2;
-		}
+        // Compute bit‑width; For val==0, this yields len==1
+        uint64_t tmp = val;
+        do {
+            len += 1;
+            tmp /= 2;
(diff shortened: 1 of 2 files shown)


weiren2:
Thanks for the feedback. I'll check `config.h.in` to understand the required changes there, and make sure the indentation matches the existing style. As suggested, I can refactor this into two commits.

caryr:
Can you check if this has been resolved using the latest code from master.

I made some changes for arm64 and these look to be mostly the same, though I cleaned things up in other places so if my changes are good enough then we can drop what you have here.

PR #1282, opened 2025-10-17
MSYS2 clang support

martinwhitaker:
Attempting to build Icarus Verilog using the clang/LLVM toolchain in MSYS2 fails because the clang/LLVM variants of dlltool and ld are not completely compatible with the binutils variants. This PR aims to work round these issues.

As a first step, the use of dlltool can be eliminated because both variants of ld support reading the .def file and creating the import library directly.

As a bonus, this PR also improves the Perl test scripts to reduce the test time. This is particularly noticeable when running in MSYS2.

### DIFF
--- a/.github/test.sh
+++ b/.github/test.sh
@@ -9,7 +9,7 @@ status=0
 
 perl vvp_reg.pl || status=1
 
-perl vpi_reg.pl || status=1
+perl vpi_reg.pl --with-pli1 || status=1
 
 python3 vvp_reg.py || status=1
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       run: |
         export PATH="/usr/local/opt/bison/bin:$PATH"
         autoconf
-        ./configure
+        ./configure --enable-libveriuser
         make -j$(nproc) check
         sudo make install
 
@@ -59,7 +59,7 @@ jobs:
     - name: Build, check and install
       run: |
         autoconf
-        ./configure
+        ./configure --enable-libveriuser
         make -j$(nproc) check
         sudo make install
 
--- a/msys2/PKGBUILD
+++ b/msys2/PKGBUILD
@@ -21,6 +21,7 @@ build() {
   cd "${srcdir}"/../..
   sh autoconf.sh
   ./configure \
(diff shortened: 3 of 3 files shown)


caryr:
Busy weekend so just reviewing now. It all looks good. One thing is should we keep the MSYS build file so it doesn't normally build with the PLI library and only does it for the CI? I'm not sure if we can pass options, etc.

caryr:
Cygwin and Ubuntu are all clean using gcc and clang. GhostBSD is clean with clang. I need to build Cygwin with the PLI library to make sure the clang build works as expected with the PLI. My guess is it should based on your linking changes.

We also need to add documentation regarding the flag to enable PLI support.

caryr:
The MSYS2 problem turned out to be with the perl version. You can install a base perl or a version dependent perl (ucr64/mingw64/clang64). I had originally installed the version dependent perl. I removed these and just install the base perl and now things are working as expected.

caryr:
I thought the default for everything was to not include PLI support. I didn't check if the C program used for windows was also updated. I just pulled and built so I would assume this automatically turned off PLI support. We should double check this the C program. I think it uses the same variables, so should be correct.

Actually, your changes and the version independent perl (msys2) is needed for things to work properly. With the version dependent perl your code needed to be further enhanced to save and restore the STDOUT and STDERR since it appears to be using the fork emulation or at least an older version of the emulation and becomes a show stopper for clang64 since it doesn't work at all. I believe this is because ithreads is not compiled in for the fork emulation in the clang64 version. I think to make things work properly we have to use the msys2 perl version. Which does seem to be faster since the CI is completing quicker now. It could just be that because the msys2 perl is a newer version the fork emulation is faster than the older version dependent perl. I'm not sure if system() would be faster in the newer version. Maybe that needs some testing.

Cygwin also seems to be faster using fork(). Which I think is likely expected since it is known to be slow spawning shells. Not a problem for general work, but definitely slower when testing.

It is interesting that the python code does seem to run faster in general, but it's hard to compare since we don't have the same tests in both.

caryr:
I think only mingw32 is deprecated and we dropped it a while ago. Though I have not read the details in a while.

caryr:
Is there a reason you are not testing the clang version with PLI?

PR #1303, opened 2026-03-05
Rename manpage to iverilog.1 to match executable name

aelmahmoudy:


caryr (approved):
I updated the name to be iverilog-docs

PR #1310, opened 2026-03-17
vvp: create libvvp as versioned library

### LINKED ISSUE #1319: vvp shared library missing SONAME
When I build iverilog with `--enable-vvp` configure option it installs libvvp.so without SONAME

rhabacker:
With these changes, the vvp library can be compiled as a shared library on Unix-like operating systems and on Windows using the MINGW compiler.

The build support for Windows has been tested with the openSUSE MinGW project (see https://build.opensuse.org/project/show/windows:mingw:win64)

Since this library is required by ngspice, a merge request was created at (https://sourceforge.net/p/ngspice/ngspice/merge-requests/49/) to allow the library to be used directly, which simplifies the packaging process. The patch was built and tested on openSUSE Leap with gcc and on the openSUSE MinGW project with a gcc cross-compiler for Windows. 

Fixes #1319 

Depends on #1328,  #1329, #1331, #1339, #1340

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,9 +14,9 @@ jobs:
   mac:
     strategy:
       fail-fast: false
-    # matrix:
-    #   libvvp: [false, true]
-    #   suffix: [false, true]
+      matrix:
+        libvvp: [true]
+        suffix: [true]
     runs-on: macos-15-intel
     name: 🍏 macOS${{ matrix.libvvp && ' +libvvp' || '' }}${{ matrix.suffix && ' +suffix' || '' }}
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,8 +14,11 @@ jobs:
   mac:
     strategy:
       fail-fast: false
+    # matrix:
+    #   libvvp: [false, true]
+    #   suffix: [false, true]
     runs-on: macos-15-intel
-    name: '🍏 macOS'
+    name: 🍏 macOS${{ matrix.libvvp && ' +libvvp' || '' }}${{ matrix.suffix && ' +suffix' || '' }}
     steps:
 
     - uses: actions/checkout@v6
@@ -27,8 +30,15 @@ jobs:
     - name: Build, check and install
       run: |
         export PATH="/usr/local/opt/bison/bin:$PATH"
+        CONFIG_OPTS="--enable-libveriuser"
+        if [ "${{ matrix.libvvp }}" = "true" ]; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-libvvp"
+        fi
+        if [ "${{ matrix.suffix }}" = "true" ]; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-suffix"
+        fi
         autoconf
-        ./configure --enable-libveriuser
+        ./configure $CONFIG_OPTS
         make -j$(nproc) check
         sudo make install
 
@@ -41,12 +51,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [
-          '22.04',
-          '24.04'
-        ]
+        os: ['22.04', '24.04']
+        # libvvp: [false, true]
+        # suffix: [false, true]
     runs-on: ubuntu-${{ matrix.os }}
-    name: '🐧 Ubuntu ${{ matrix.os }}'
+    name: 🐧 Ubuntu ${{ matrix.os }}${{ matrix.libvvp && ' +libvvp' || '' }}${{ matrix.suffix && ' +suffix' || '' }}
     steps:
 
     - uses: actions/checkout@v6
@@ -62,8 +71,15 @@ jobs:
 
     - name: Build, check and install
       run: |
+        CONFIG_OPTS="--enable-libveriuser"
+        if [ "${{ matrix.libvvp }}" = "true" ]; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-libvvp"
+        fi
+        if [ "${{ matrix.suffix }}" = "true" ]; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-suffix"
+        fi
         autoconf
-        ./configure --enable-libveriuser
+        ./configure $CONFIG_OPTS
         make -j$(nproc) check
         sudo make install
 
@@ -82,11 +98,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        msystem: [MINGW64, UCRT64, CLANG64]
+        # libvvp: [false, true]
+        # suffix: [false, true]
         include:
           - { msystem: MINGW64, env: x86_64 }
           - { msystem: UCRT64,  env: ucrt-x86_64 }
           - { msystem: CLANG64, env: clang-x86_64 }
-    name: 🟪 ${{ matrix.msystem}}
+    name: 🟪 ${{ matrix.msystem }}${{ matrix.libvvp && ' +libvvp' || '' }}${{ matrix.suffix && ' +suffix' || '' }}
     defaults:
       run:
         shell: msys2 {0}
@@ -116,9 +135,17 @@ jobs:
     - name: Build and check
       run: |
         cd msys2
+        CONFIG_OPTS=""
         if [ ${{ matrix.msystem }} != "CLANG64" ] ; then
-            export IVL_CONFIG_OPTIONS="--enable-libveriuser"
+            CONFIG_OPTS="$CONFIG_OPTS --enable-libveriuser"
         fi
+        if [ "${{ matrix.libvvp }}" = "true" ] ; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-libvvp"
+        fi
+        if [ "${{ matrix.suffix }}" = "true" ]; then
+            CONFIG_OPTS="$CONFIG_OPTS --enable-suffix"
+        fi
+        export IVL_CONFIG_OPTIONS="$CONFIG_OPTS"
         makepkg-mingw --noconfirm --noprogressbar -sCLf
 
     - name: Install
@@ -130,5 +157,5 @@ jobs:
 
     - uses: actions/upload-artifact@v7
       with:
-        name: ${{ matrix.msystem }}
+        name: 🟪 ${{ matrix.msystem }}${{ matrix.libvvp && ' +libvvp' || '' }}
         path: msys2/*.zst
--- a/Makefile.in
+++ b/Makefile.in
@@ -145,20 +145,8 @@ endif
 check: all
 	$(foreach dir,$(SUBDIRS),$(MAKE) -C $(dir) $@ && ) true
 	test -r check.conf || cp $(srcdir)/check.conf .
-	driver/iverilog -B. -BMvpi -BPivlpp -tcheck -ocheck.vvp $(srcdir)/examples/hello.vl
-ifeq (@WIN32@,yes)
-ifeq (@install_suffix@,)
-	vvp/vvp -M- -M./vpi ./check.vvp | grep 'Hello, World'
-else
-	# On Windows if we have a suffix we must run the vvp part of
-	# the test with a suffix since it was built/linked that way.
-	ln vvp/vvp.exe vvp/vvp$(suffix).exe
-	vvp/vvp$(suffix) -M- -M./vpi ./check.vvp | grep 'Hello, World'
-	rm vvp/vvp$(suffix).exe
-endif
-else
-	$(ENV_VVP) vvp/vvp -M- -M./vpi ./check.vvp | grep 'Hello, World'
-endif
+	driver/iverilog@EXEEXT@ -B. -BMvpi -BPivlpp -tcheck -ocheck.vvp $(srcdir)/examples/hello.vl
+	$(ENV_VVP) vvp/vvp$(suffix)@EXEEXT@ -M- -M./vpi ./check.vvp | grep 'Hello, World'
 
 check-installed check-installed-vpi check-installed-vvp check-installed-vvp-py:
 	$(MAKE) -C ivtest $@
--- a/configure.ac
+++ b/configure.ac
@@ -7,6 +7,9 @@ m4_define([VER_MAJOR], [14])
 m4_define([VER_MINOR], [0])
 m4_define([VER_EXTRA], [devel])
 
+dnl define libvvp ABI version
+m4_define([LIBVVP_SOVERSION], [1])
+
 AC_INIT([iverilog], [VER_MAJOR.VER_MINOR (VER_EXTRA)])
 AC_SUBST([VERSION_MAJOR], [VER_MAJOR])
 AC_SUBST([VERSION_MINOR], [VER_MINOR])
@@ -14,6 +17,11 @@ AC_SUBST([VERSION_EXTRA], [" (VER_EXTRA)"])
 AC_SUBST([VERSION], ["VER_MAJOR.VER_MINOR (VER_EXTRA)"])
 # used in res.rc
 AC_SUBST([PRODUCTVERSION], ["VER_MAJOR,VER_MINOR,0,0"])
+
+# setup libvvp soversion, which depends on abi not package version
+AC_SUBST([LIBVVP_SOVERSION], [LIBVVP_SOVERSION])
+# setup libvvp version
+AC_SUBST([LIBVVP_VERSION], [LIBVVP_SOVERSION.VER_MAJOR.VER_MINOR])
 
 AC_CONFIG_SRCDIR([netlist.h])
 # Need a stamp file like the other header files
@@ -209,7 +217,14 @@ AC_FUNC_FSEEKO
 # Build VVP as a library and stub
 AC_ARG_ENABLE([libvvp],
     [AS_HELP_STRING([--enable-libvvp], [build VVP as a shared library])],
-    [AC_SUBST(LIBVVP, yes)],[])
+    [enable_libvvp=yes],
+    [enable_libvvp=no])
+
+AC_SUBST([LIBVVP], [$enable_libvvp])
+
+AS_IF([test "x$enable_libvvp" = "xyes"],
+    [AC_MSG_NOTICE([Building with libvvp support enabled])],
+    [AC_MSG_NOTICE([Building with libvvp support disabled])])
 
 AC_ARG_ENABLE([libveriuser],
     [AS_HELP_STRING([--enable-libveriuser], [include support for PLI 1 (deprecated)])],
@@ -434,6 +449,7 @@ AC_CONFIG_FILES([
     vhdlpp/Makefile
     vpi/Makefile
     vvp/Makefile
+    vvp/libvvp.pc
     vvp/vvp.man
 ])
 AC_OUTPUT
--- a/vvp/Makefile.in
+++ b/vvp/Makefile.in
@@ -61,7 +61,7 @@ CPPFLAGS = $(INCLUDE_PATH) @CPPFLAGS@ @DEFS@ -DICARUS_VPI_CONST=const
 CFLAGS = @WARNING_FLAGS@ @WARNING_FLAGS_CC@ @CFLAGS@
 CXXFLAGS = @WARNING_FLAGS@ @WARNING_FLAGS_CXX@ @CXXFLAGS@
 LDFLAGS = @rdynamic@ @LDFLAGS@
-LIBS = @LIBS@ @EXTRALIBS@
+LIBS = @LIBS@ @EXTRALIBS@ @DLLIB@
 
 ifeq (@WIN32@,yes)
 SLDIR=$(bindir)
@@ -70,8 +70,6 @@ else
 SLDIR=$(libdir)
 SLEXT=so
 endif
-
-dllib=@DLLIB@
 
 MDIR1 = -DMODULE_DIR1='"$(libdir)/ivl$(suffix)"'
 
@@ -153,10 +151,10 @@ ifeq (@WIN32@,yes)
 # cocotb to build VPI modules without using our vpi_user.h and libvpi.a.
 # The .def file controls what is exported.
 vvp@EXEEXT@: main.o $O $(srcdir)/vvp.def
-	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ $(LDFLAGS) $(srcdir)/vvp.def main.o $O $(dllib) $(LIBS)
+	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ $(LDFLAGS) $(srcdir)/vvp.def main.o $O $(LIBS)
 else
 vvp@EXEEXT@: $O main.o
-	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ main.o $O $(LIBS) $(dllib)
+	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ main.o $O $(LIBS)
 endif
 endif
 
--- a/vvp/Makefile.in
+++ b/vvp/Makefile.in
@@ -73,22 +73,24 @@ endif
 
 MDIR1 = -DMODULE_DIR1='"$(libdir)/ivl$(suffix)"'
 
-VPI = vpi_modules.o vpi_bit.o vpi_callback.o vpi_cobject.o vpi_const.o vpi_darray.o \
-      vpi_event.o vpi_iter.o vpi_mcd.o \
-      vpi_priv.o vpi_scope.o vpi_real.o vpi_signal.o vpi_string.o vpi_tasks.o vpi_time.o \
-      vpi_vthr_vector.o vpip_bin.o vpip_hex.o vpip_oct.o \
-      vpip_to_dec.o vpip_format.o vvp_vpi.o
+CORE_OBJ = lib_main.o \
+           parse.o parse_misc.o lexor.o arith.o array_common.o array.o bufif.o compile.o \
+           concat.o dff.o class_type.o enum_type.o extend.o file_line.o latch.o npmos.o part.o \
+           permaheap.o reduce.o resolv.o \
+           sfunc.o stop.o \
+           substitute.o \
+           symbols.o ufunc.o codes.o vthread.o schedule.o \
+           statistics.o tables.o udp.o vvp_island.o vvp_net.o vvp_net_sig.o \
+           vvp_object.o vvp_cobject.o vvp_darray.o event.o logic.o delay.o \
+           words.o island_tran.o
 
-O = lib_main.o \
-    parse.o parse_misc.o lexor.o arith.o array_common.o array.o bufif.o compile.o \
-    concat.o dff.o class_type.o enum_type.o extend.o file_line.o latch.o npmos.o part.o \
-    permaheap.o reduce.o resolv.o \
-    sfunc.o stop.o \
-    substitute.o \
-    symbols.o ufunc.o codes.o vthread.o schedule.o \
-    statistics.o tables.o udp.o vvp_island.o vvp_net.o vvp_net_sig.o \
-    vvp_object.o vvp_cobject.o vvp_darray.o event.o logic.o delay.o \
-    words.o island_tran.o $(VPI)
+VPI_OBJ = vpi_modules.o vpi_bit.o vpi_callback.o vpi_cobject.o vpi_const.o vpi_darray.o \
+          vpi_event.o vpi_iter.o vpi_mcd.o \
+          vpi_priv.o vpi_scope.o vpi_real.o vpi_signal.o vpi_string.o vpi_tasks.o vpi_time.o \
+          vpi_vthr_vector.o vpip_bin.o vpip_hex.o vpip_oct.o \
+          vpip_to_dec.o vpip_format.o vvp_vpi.o
+
+LIB_OBJ = $(CORE_OBJ) $(VPI_OBJ)
 
 all: dep vvp@EXEEXT@ vvp.man
 
@@ -115,7 +117,7 @@ distclean: clean
 	rm -f Makefile config.log
 	rm -f stamp-config-h config.h
 
-cppcheck: $(O:.o=.cc) draw_tt.c
+cppcheck: $(LIB_OBJ:.o=.cc) draw_tt.c
 	cppcheck --enable=all --std=c99 --std=c++11 -f \
 	         --check-level=exhaustive \
 	         --suppressions-list=$(srcdir)/../cppcheck-global.sup \
@@ -143,18 +145,18 @@ CPPFLAGS+= -fpic
 vvp@EXEEXT@: main.o $(srcdir)/vvp.def libvvp$(suffix).$(SLEXT)
 	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ main.o -L. $(LDFLAGS) -lvvp$(suffix) $(LIBS)
 
-libvvp$(suffix).$(SLEXT): $O
-	$(CXX) -shared $(LDFLAGS) -o libvvp$(suffix).$(SLEXT) $O $(LIBS) $(dllib)
+libvvp$(suffix).$(SLEXT): $LIB_OBJ
+	$(CXX) -shared $(LDFLAGS) -o libvvp$(suffix).$(SLEXT) $LIB_OBJ $(LIBS) $(dllib)
 else
 ifeq (@WIN32@,yes)
 # To support cocotb, we export the VPI functions directly. This allows
 # cocotb to build VPI modules without using our vpi_user.h and libvpi.a.
 # The .def file controls what is exported.
-vvp@EXEEXT@: main.o $O $(srcdir)/vvp.def
-	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ $(LDFLAGS) $(srcdir)/vvp.def main.o $O $(LIBS)
+vvp@EXEEXT@: main.o $LIB_OBJ $(srcdir)/vvp.def
+	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ $(LDFLAGS) $(srcdir)/vvp.def main.o $LIB_OBJ $(LIBS)
 else
-vvp@EXEEXT@: $O main.o
-	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ main.o $O $(LIBS)
+vvp@EXEEXT@: $LIB_OBJ main.o
+	$(CXX) $(LDFLAGS) -o vvp@EXEEXT@ main.o $LIB_OBJ $(LIBS)
 endif
 endif
 
@@ -243,4 +245,4 @@ ifeq (@LIBVVP@,yes)
 	rm -f "$(DESTDIR)$(ivl_includedir)/libvvp.h"
 endif
 
--include $(patsubst %.o, dep/%.d, $O)
+-include $(patsubst %.o, dep/%.d, $LIB_OBJ)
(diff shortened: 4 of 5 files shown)


caryr (changes requested):
I need to look at this in more detail, but the first thing I noticed is I think the library SO version should match the program version and that should be determined using the `version` program.

Also what systems have you validated this on. This is where we often have breakage because of differences in compilers/OSs.

rhabacker:
https://github.com/steveicarus/iverilog/commit/6165f77f8621ea5cc3e9ecfbbce45a57c24a3007 fixes issues with installing the pkgconfig file during cross-compilation for Mingw. It should only be installed when using ./configure --enable-libvvp and should take DESTDIR into account during installation.

martinwhitaker:
The issue with making the library version independent of the compiler version is that the vvp input file API is allowed to change between major versions. So even when the libvvp client ABI hasn't changed, a new version will not necessarily work with previously compiled simulation code. I would suggest making the library major version reflect the client ABI, and making the library minor and extra version match the compiler major and minor version. Then clients can decide for themselves how much compatibility they want to enforce.

On a quick look at your patch, I can see changes to the makefile install target, but no corresponding changes to the uninstall target.

Changing the CI to test the libvvp build means we are no longer testing the non-libvvp build, which is the normal use case.

caryr:
I've been thinking about this specific issue the last few days and I agree with Martin. People may not know this but there is code in the compiler and run-time that make sure the user compiles and runs code that is compatible. While the ABI for using the library may not change which is what the library version is primarily concerned with and I agree with the NGSPICE folks it does not need to match the Icarus version. We also need the version in the name somewhere to help the user make sure they are using compatible version.

See: `:ivl_version "14.0 (devel)" ...`

Currently 13.0 and 14.0 are compatible if you hack the version, but there will be changes in 14 that will eventually cause it to diverge from 13.0. One of the requirements in a stable branch (e.g. 13) is that we do not break the syntax between the compiler and run-time. The fundamental rule is that the major versions must match between the compiler and the run-time and a warning is generated if the minor version of the compiler is greater than the run-time version indicating that there could be compatibility issues.

Maybe maybe making a 1.13.0 for stable and 1.14.0 for our current development is the best way to go.

rhabacker:
Updated patch with uninstalling library files added.

caryr:
Actually one problem with making it 1.<ivl_version> is you cannot have V13 and V14 installed at the same time. Which may not be an issue for general users, but does have an impact on development testing. The advantage for general users is they can upgrade the the Icarus version without needing to recompile other tools (like NGSPICE).

rhabacker:
Updated patches
- Only add generated files to configure.ac when covered by the patch
- Fixed not uninstalling pkgconfig file
- Fixed library prefix for non MinGW compilers on Windows
- CI: Added additional linux job for compiling with libvvp

rhabacker:
Update
- squashed fixup commit for lib prefix into versioned vpp library support commit
- Fix unrelated install bug fix

caryr:
I've been busy so have not looked at your recent changes. I'll try to get back to this soon.

Can you also look at #1313 and #1319 to make sure your changes also address these recent issue submissions. For #1313 you may need to run uninstall to make sure there is no library already installed.

rhabacker:
> Can you also look at #1313 

This is already covered by commit [vvp: build and install libvvp as a versioned shared library](https://github.com/steveicarus/iverilog/pull/1310/commits/cee2126bf4c80cfeb39a139a4b9acd13dae5b485) for `in-source` builds.

> #1319 to make sure your changes also address these recent issue submissions. 

Adding SONAME to the shared library is also added by the commit mentioned above.

rhabacker:
Update
- Use autoconf to generated man files instead of using a sub process
- Use autoconf variables to create pdf filenames
- Squashed fixup commits

> This workflow requires approval from a maintainer
@caryr: Please approve this, as it will ensure that any potential issues with this pull request are identified.

caryr:
I approved the workflow. The first thing I noticed is there is no indication which of the Linux builds is with/without the shared library, even looking at the autoconf part of the logs it's not obvious. We should have something that make it clear somewhere in there that we are building a shared library version. I'll try to review this in more detail soon. I've been out of work sick so what little energy I do have is going to keep up there and I don't have much extra for Icarus. Hopefully soon.

rhabacker:
Update
- Extended CI job name building with --enable-libvvp
- Simplified build rules for shared library
- Fixed out of source builds for generated man files

caryr:
I appreciate limiting the resources used for the library built version in the CI flow, but I think we should run some tests to verify the build after installation perl/python vvp and VPI in the CI. I can help or add this if it's too much for you. I have some ideas.

caryr:
Did I miss or is it still missing that when running autoconf it mentions that the build is with/without the libvvp option?

caryr:
Maybe something like was done for valgrind?

rhabacker:
> Did I miss or is it still missing that when running autoconf it mentions that the build is with/without the libvvp option?

```
git log configure
commit 43725602904228e554e82cad51776050b06e6c0b
Author: Martin Whitaker <icarus@martin-whitaker.me.uk>
Date:   Sun Oct 12 11:32:34 2025 +0100

    Post-snapshot cleanup

commit 3e7cc4eac241e3649fc89e82d1ef02c561623d45 (tag: s20251012)
Author: Martin Whitaker <icarus@martin-whitaker.me.uk>
Date:   Sun Oct 12 11:32:34 2025 +0100

    Creating snapshot s20251012
```
The configuration file is updated only by commits, not by running `sh autoconf.sh`, which is required when changes are made to `configure.ac` and is missing in CI.

rhabacker:
> > Did I miss or is it still missing that when running autoconf it mentions that the build is with/without the libvvp option?
> 
> ```
> git log configure
> commit 43725602904228e554e82cad51776050b06e6c0b
> Author: Martin Whitaker <icarus@martin-whitaker.me.uk>
> Date:   Sun Oct 12 11:32:34 2025 +0100
> 
>     Post-snapshot cleanup
> 
> commit 3e7cc4eac241e3649fc89e82d1ef02c561623d45 (tag: s20251012)
> Author: Martin Whitaker <icarus@martin-whitaker.me.uk>
> Date:   Sun Oct 12 11:32:34 2025 +0100
> 
>     Creating snapshot s20251012
> ```
> 
> The configuration file is updated only by commits, not by running `sh autoconf.sh`, which is required when changes are made to `configure.ac` and is missing in CI.

I'm wrong. In the CI `autoconf` is called,  which should create a new configure file, so the root cause must be different.

rhabacker:
> I'm wrong. In the CI `autoconf` is called, which should create a new configure file, so the root cause must be different.

When I run on this branch checked out:
```
sh autoconf.sh
./configure --enable-libvvp
make
rm -rf $PWD/tmp
make install DESTDIR=$PWD/tmp
ls $PWD/tmp/usr/local/lib
ivl  libvpi.a  libvvp.so  libvvp.so.1  libvvp.so.14.0  pkgconfig
```
I get the expected results.

rhabacker:
> Maybe something like was done for valgrind?

When running make check a test using vpp an libvpp is already available: 
```
make check
...
test -r check.conf || cp ../iverilog/check.conf .
driver/iverilog -B. -BMvpi -BPivlpp -tcheck -ocheck.vvp ../iverilog/examples/hello.vl
LD_LIBRARY_PATH=../iverilog/vvp vvp/vvp -M- -M./vpi ./check.vvp | grep 'Hello, World'
Hello, World
```
You you have something else in mind ?

rhabacker:
Updated branch
- rebased
- fixed generating version_base.h and config.h with out of source builds

rhabacker:
Updated branch
- added additional test stage to CI job `lin-with-libvvp`

caryr:
The tests failed for the libvvp versions previously. Do you need to update/add LD_LIBRARY for the test script?

caryr:
Also, has all the documentation been updated to reflect the change in where the version information is contained? I want to try reviewing this in detail over the weekend and push it if possible.

rhabacker:
> Also, has all the documentation been updated to reflect the change in where the version information is contained? I want to try reviewing this in detail over the weekend and push it if possible.

See "[Update documentation to clarify that “version.exe” is no longer used](https://github.com/steveicarus/iverilog/pull/1310/changes/699d58df5eebbd65853c8264511460181e5fa054)"

rhabacker:
> The tests failed for the libvvp versions previously. Do you need to update/add LD_LIBRARY for the test script?

Yes, LD_LIBRARY_PATH is required here as well. 

In my opinion, however, it would be better to control this via a Makefile to also support builds outside the source directory, since the necessary directories are already defined in the build system. For example, you could add a target named `full-check` or something similar and call it in the CI script as `make full-check`.

caryr:
My idea for full check is it runs all the tests before installing. Which requires the local paths not the installed paths along with some other magic. Maybe this should be check-installed and we save check-full for checking the full test suite before installing? How does that sound?

All the test scripts use the installed version so in/out of source builds should not matter for them. They just need the installed path.

caryr:
Do we need to mention that the user may need to updater their LD_LIBRARY_PATH to get the library version to work properly even after being installed (e.g. in /usr/local/...).

rhabacker:
> My idea for full check is it runs all the tests before installing. Which requires the local paths not the installed paths along with some other magic. Maybe this should be check-installed and we save check-full for checking the full test suite before installing? How does that sound?

Something like this? 
```
make check                # basic tests (build tree) (-> target “actual check”)
make check-full         # all tests (build tree, stricter) (calls .github/test.sh)
make install
make check-installed   # tests based on the installed binaries (what should be called here)
```

> make check-full         # all tests (build tree, stricter) (calls .github/test.sh)

Regarding the target name—in .github/test.sh it says 
```
echo “Using the bundled ivtest to run regression tests.”
```
Are only regression tests run here? 

and
```
cd ivtest
```
It looks to me as though this script could be moved to a Makefile.in in the ivtest directory to simplify things.

caryr:
ivtests used to be separate from iverilog. The github test scrip checks using the installed version. We should likely update the script to take the suffix and if a LIBRARY path is needed, etc. The default is no suffix and no LIBRARY needed. make check-installed can pass these from the Makefile when called. Make check-full needs the things from make check to run using the uninstalled version, but applied to the complete regression. Locally I run a lot more check options, but for the basic user checking the perl/python VVP and perl VPI tests are what is required, which matches the github script.

I have a more comprehensive script that extracts the failed tests, but it can currently only do that for the perl VVP tests since that script is the only one that generates a log file. Something I was planning to fix for the 14.0 release, but that has been pushed to a post 14 task and still needs to be completed. I would like to eventually replace the current script with my updated script.

aelmahmoudy:
I had to change `LD_LIBRARY_PATH` in `vvp/Makefile.in` to:
`LD_LIBRARY_PATH=$(builddir)/..:$(builddir)` to get `make check` pass

rhabacker:
Update
- Commits merged
- Uses an absolute file path for the build directory specified in LD_LIBRARY_PATH to avoid incorrect calculations caused by relative paths
- added `make check-full` target

rhabacker:
Update
- ivtest: Added additional targets to fix the issue with (un)installation

Note: The “ivtest” directory contains additional targets for running partial tests; these can be executed with `make -C ivtest check-vvp|check-vpi|check-py`

caryr:
I know I asked this earlier, but can I get a confirmation this PR should be marked as closing the issues reported in #1313 and #1319?

rhabacker:
> I know I asked this earlier, but can I get a confirmation this PR should be marked as closing the issues reported in #1313 and #1319?

Yes, I added these issues to be closed by this pr.

caryr:
A few issues I found while compiling on Ubuntu 22.04 (both gcc and clang)/GhostBSD(clang):

1. If you run make check-full before installing all the tests in check-vpi fail. They run properly after installing.
2. CREATE_RELEASE.sh needs to be updated to work with the new version scheme.
3. Make distclean does not cleanup some of the new files.

And instead of accepting --enable-libvvp we should fail during configure for systems that are not currently supported (e.g. msys2, ...).

caryr:
msys2 using gcc is working properly generating a DLL, including check-vpi. It's just the clang versions that fails to create a DLL properly.

rhabacker:
Update
- Apply fixes to scripts/CREATE_BRANCH.sh and scripts/CREATE_RELEASE.sh
- Fix removing generated files with `make distclean`
- CI: Add jobs with libvvp enabled for all platforms

rhabacker:
Update
- Fix building on MacOSX
- Fix building with Clang on Windows

rhabacker:
Update
-  CI: Fix building all variants on Windows, not only with clang

caryr:
I need to test more systems and review a bit more. Some issues so far:

1. make distclean is not cleaning up and removing  the new ivtest directory/Makefile. Make sure ivtest is removed like the other directories since for an out of repo build it is removed, but for an in repo build only the Makefile should be removed.
2. check-vpi is still failing if nothing has been installed. It looks like it cannot find the header files. You need to run make uninstall first to make sure nothing is still installed from a previous attempt to reproduce the issue. You could also check this by skipping the install step in the CI script before starting the testing.
3. As you can see macOS is failing when libvvp is enabled.
4. Can you rename check-py in the ivtest/Makefile to be check-vvp-py. I plan to add more targets once this is merged and using that naming convention will be more compatible.

I'll check on more systems when I have time.

caryr:
The make check-full results I've reported so far are mostly bogus. It looks like it is using the default (no suffix) install on the machine or some weird hybrid. The suffix is not being used in/passed to the ivtest Makefile.

I removed all the installed versions for cygwin and then tried the library versions. They completely fail for both the pre install and post install cases using make check-full (as expected since the suffix is wrong). After being installed all the vvp, vvp-py and vpi tests run as expected when using my normal scripts with the correct suffix. I can't check pre installed since that is setup by the Makefile which does not pass the suffix. I checked using both gcc and clang and they ran as expected.

rhabacker:
Update
- Attempt to fix finding libvvp on MacOSX
- let `make check-vpi` be more verbose
- let `make check_vpi|check-vvp|check-vvp-py` be callable from the top build dir
- Renamed make target from `check-py` to `check-vvp-py`
- Fix `make distclean` in ivtest sub directory
- `make clean`: remove generated vpi files in ivtest sub directory

rhabacker:
Running, for example, `make check-vvp` in a non-installed environment fails because hard-coded paths are used that do not match the structure of the build directory.

The actual command line used on Linux is 
```
LD_LIBRARY_PATH=/home/user/src/iverilog/vvp PATH=/home/user/src/iverilog/vvp:/usr/bin:/bin perl vvp_reg.pl

==> Failed – Execution of iverilog, see “log/macro_with_args.log” for details.
```
The log file shows
```
$ cat ivtest/log/macro_with_args.log 
sh: /home/user/src/iverilog/lib64/ivl/ivlpp: File or directory not found
sh: /home/user/src/iverilog/lib64/ivl/ivl: File or directory not found
```
By adding additional command-line options and a new support option `-B I<path-to-ivl>`, I can get some results: 
```
PATH=/home/user/src/iverilog/vvp:/home/user/src/iverilog/driver:$PATH iverilog -v -B /home/user/src/iverilog/vpi -B I/home/user/src/iverilog -B P/home/user/src/iverilog/ivlpp -o vsim -D__ICARUS_UNSIZED__  ./ivltests/br_gh515.v
```
which is translated into:
```
translate: /home/user/src/iverilog/ivlpp/ivlpp -v -L -F"/tmp/ivrlg23a0c2f58" -f"/tmp/ivrlg3a0c2f58" -p"/tmp/ivrli3a0c2f58" |/home/user/src/iverilog/ivl -v -C"/tmp/ivrlh3a0c2f58" -C"/home/user/src/iverilog/vpi/vvp.conf" -- -
ERROR: Unable to read config file: /home/ralf/src/iverilog/vpi/vvp.conf
Using language generation: IEEE1364-2005,no-specify,no-interconnect,xtypes,icarus-misc
PARSING INPUT
LOCATING TOP-LEVEL MODULES
   test
 ... done, 0 seconds.
ELABORATING DESIGN
 ... done, 0 seconds.
RUNNING FUNCTORS
CALCULATING ISLANDS
 ... done, 0 seconds.
CODE GENERATION
: error: target_design entry point is missing.
error: Code generator failure: -2
```
but there are still errors caused by not finding vvp.conf, which is located in another directory:
```
cd ~/src/iverilog
$ find -name vvp.conf
tgt-vvp/vvp.conf
```
In addition, further adjustments are required, for example finding `flag:DLL=vvp.tgt` and `flag:VVP_EXECUTABLE=/usr/local/bin/vvp`. 

This raises the question of whether the transition from the current installation base to one that uses the current build system goes beyond the scope of this pull request.

rhabacker:
Update
- squashed CI related commits

rhabacker:
> Actually one problem with making it 1.<ivl_version> is you cannot have V13 and V14 installed at the same time. Which may not be an issue for general users, but does have an impact on development testing. 

This problem can be avoided on Unix-like operating systems by using `-Wl,-rpath`. If a development version is installed under `/usr/local` and the installed executables are linked with `-Wl,-rpath`, they will access the shared libraries from the same installation root directory, unless this is overridden by using LD_LIBRARY_PATH.
If applications from a regular installation under /usr are also linked with rpath, they will likewise use the shared libraries from /usr/lib*.

caryr:
I'm perfectly OK with this being something that can only be called post install. I was confused previously because it was working, but that was not real since it was calling the already installed non-suffix version. I assume we want the tests to use the suffix that matches what was used to build the executable.

Maybe the suffix fixes the library versioning since only one version at a time will be installed as the base library all the rest will have a suffix to distinguish them from each other. For example I have the non-suffix version for the development gcc, no library, `-cl` for the clang compiled, no library version, `-13` for the latest release. I am using `-gso` and `-cso` for testing the library versions.

I think that implies we can link `ngspice` to the default non-suffix version and as that is upgraded the library and compiler will be kept in sync by the normal install rules, but `ngspice` can automatically link against whatever is the latest unless the actual library interface is changed. Maybe we need to review that interface and make sure it is functionally complete so we have something stable.

caryr:
If the new checking is only going to be for after it is installed then I think the target should be renamed to make check-installed and we reserve check-full for running the test suite before it is installed. henever we get around to getting that to work. Thoughts?

ETA. You could keep the check-full target and have it print a message that we do not currently have support to run the full test suite before it is installed and to use check-installed after Icarus has been installed.

rhabacker:
Update
- Attempt to fix build/install issues when using `configure --enable-suffix`

rhabacker:
Update
- Fix running installed tests on Linux platform with libvvp enabled

rhabacker:
> Maybe we need to review that interface and make sure it is functionally complete so we have something stable.

Here are a few hints:

1. These functions are imported by the vvp executable from the shared libvvp  library
```
 objdump -x vvp.exe | less
        DLL Name: libvvp-1.dll
        vma:     Ordinal  Hint  Member-Name  Bound-To
        000113b8  <none>  0089  _Z13schedule_stopi  # c++ symbol
        000113c0  <none>  0fd4  vpi_chk_error
        000113c8  <none>  0fd5  vpi_compare_objects
        000113d0  <none>  0fd6  vpi_control
        000113d8  <none>  0fd7  vpi_flush
        000113e0  <none>  0fd8  vpi_fopen
        000113e8  <none>  0fd9  vpi_free_object
        000113f0  <none>  0fda  vpi_get
        000113f8  <none>  0fdb  vpi_get64
        00011400  <none>  0fdc  vpi_get_delays
        00011408  <none>  0fdd  vpi_get_file
        00011410  <none>  0fde  vpi_get_str
        00011418  <none>  0fdf  vpi_get_systf_info
        00011420  <none>  0fe0  vpi_get_time
        00011428  <none>  0fe1  vpi_get_userdata
        00011430  <none>  0fe2  vpi_get_value
        00011438  <none>  0fe3  vpi_get_vlog_info
        00011440  <none>  0fe4  vpi_handle
        00011448  <none>  0fe5  vpi_handle_by_index
        00011450  <none>  0fe6  vpi_handle_by_name
        00011458  <none>  0fe7  vpi_handle_multi
        00011460  <none>  0fe8  vpi_iterate
        00011468  <none>  0fe9  vpi_mcd_close
        00011470  <none>  0fea  vpi_mcd_flush
        00011478  <none>  0feb  vpi_mcd_name
        00011480  <none>  0fec  vpi_mcd_open
        00011488  <none>  0fed  vpi_mcd_printf
        00011490  <none>  0fee  vpi_mcd_vprintf
        00011498  <none>  0ff0  vpi_printf
        000114a0  <none>  0ff1  vpi_put_delays
        000114a8  <none>  0ff2  vpi_put_userdata
        000114b0  <none>  0ff3  vpi_put_value
        000114b8  <none>  0ff4  vpi_register_cb
        000114c0  <none>  0ff5  vpi_register_systf
        000114c8  <none>  0ff6  vpi_release_handle
        000114d0  <none>  0ff7  vpi_remove_cb
        000114d8  <none>  0ff9  vpi_scan
        000114e0  <none>  0ffa  vpi_sim_control
        000114e8  <none>  0ffb  vpi_sim_vcontrol
        000114f0  <none>  0ffd  vpi_vprintf
        000114f8  <none>  0fff  vpip_add_module_path
        00011500  <none>  1000  vpip_calc_clog2
        00011508  <none>  1001  vpip_clear_module_paths
        00011510  <none>  1002  vpip_count_drivers
        00011518  <none>  1006  vpip_format_strength
        00011520  <none>  1007  vpip_load_module
        00011528  <none>  1008  vpip_make_systf_system_defined
        00011530  <none>  100a  vpip_mcd_rawwrite
        00011538  <none>  100b  vpip_set_return_value
        00011540  <none>  1010  vvp_init
        00011548  <none>  1012  vvp_run
        00011550  <none>  1013  vvp_set_quiet_flag
        00011558  <none>  1014  vvp_set_stop_is_finish
        00011560  <none>  1015  vvp_set_verbose_flag
```
Since libvpi also exports these functions, it looks that they are simply forwarded via libvvp-1.dll 
```
$ find -name libvpi.a | xargs nm
0000000000000760 T vpi_chk_error
00000000000007a0 T vpi_compare_objects
00000000000008a0 T vpi_control
0000000000000690 T vpi_flush
0000000000000940 T vpi_fopen
00000000000007e0 T vpi_free_object
0000000000000280 T vpi_get
00000000000002c0 T vpi_get64
0000000000000340 T vpi_get_delays
0000000000000980 T vpi_get_file
0000000000000300 T vpi_get_str
00000000000000c0 T vpi_get_systf_info
0000000000000440 T vpi_get_time
0000000000000480 T vpi_get_userdata
00000000000003c0 T vpi_get_value
0000000000000860 T vpi_get_vlog_info
00000000000001c0 T vpi_handle
0000000000000140 T vpi_handle_by_index
0000000000000100 T vpi_handle_by_name
0000000000000180 T vpi_handle_multi
0000000000000200 T vpi_iterate
0000000000000540 T vpi_mcd_close
0000000000000580 T vpi_mcd_flush
00000000000005c0 T vpi_mcd_name
0000000000000500 T vpi_mcd_open
0000000000000600 T vpi_mcd_printf
0000000000000650 T vpi_mcd_vprintf
00000000000009c0 T vpip_calc_clog2
0000000000000a00 T vpip_count_drivers
0000000000000a40 T vpip_format_strength
0000000000000a80 T vpip_make_systf_system_defined
0000000000000ac0 T vpip_mcd_rawwrite
00000000000006d0 T vpi_printf
0000000000000000 b vpip_routines
0000000000000b40 T vpip_set_callback
0000000000000b00 T vpip_set_return_value
0000000000000380 T vpi_put_delays
00000000000004c0 T vpi_put_userdata
0000000000000400 T vpi_put_value
0000000000000000 T vpi_register_cb
0000000000000080 T vpi_register_systf
0000000000000820 T vpi_release_handle
0000000000000040 T vpi_remove_cb
0000000000000240 T vpi_scan
00000000000008f0 T vpi_sim_control
0000000000000720 T vpi_vprintf
```

2. The shared library exports to much functions
```
objdump -x libvvp-1.dll | grep "Export RVA" | wc -l
4132
```
It is necessary to restrict the number of exported functions. This can be done using a .def file or by decorating the exported functions in the public header with __declspec( dllexport|dllimport ).

rhabacker:
> This problem can be avoided on Unix-like operating systems by using `-Wl,-rpath`

Fixed with https://github.com/steveicarus/iverilog/pull/1310#issuecomment-4330413961

rhabacker:
> > Can you also look at #1313
> 
> This is already covered by commit [vvp: build and install libvvp as a versioned shared library](https://github.com/steveicarus/iverilog/pull/1310/commits/cee2126bf4c80cfeb39a139a4b9acd13dae5b485) for `in-source` builds.

Although the solution to this problem is included in this PR, it is independent of it and can be released separately in advance.

rhabacker:
> > > Can you also look at #1313

> Although the solution to this problem is included in this PR, it is independent of it and can be released separately in advance.

See https://github.com/steveicarus/iverilog/pull/1329

caryr:
The other two PRs have been merged so this one can be rebased. Before we release this one we should drop the 4x increase in tests, though we should leave in the hooks to test everything when needed.

caryr:
FYI after integrating the previous PRs there appears to now be a missing config file dependency since the config files are always being regenerated.

rhabacker:
I am currently in the process of submitting commits that are independent of the integration of the versioned libvvp as separate pull requests. After that, I will rebase this branch.

caryr:
I've noticed there are now a bunch of checks like the following appearing for every invocation of make:

./config.status config.h
config.status: creating config.h
config.status: config.h is unchanged

I've not had time to look into them, but I think they started appearing after the version changes were committed.

rhabacker:
> The other two PRs have been merged so this one can be rebased. Before we release this one we should drop the 4x increase in tests, though we should leave in the hooks to test everything when needed.

Should all build variants be disabled using --enable-libvvp? And if not, for which platforms should they remain enabled?

caryr:
I was thinking enabling the suffix and libvvp would be the super set. If we also enabled no suffix or libvvp that would build the corners, but that is still a doubling of the CI usage. Maybe we should do the full test with the suffix/libvvp version and then something much simpler from a test perspective for the base build.  So build both but only fully test one of them. What do you think? I'll turn this in with support for just the corners enabled and we can tweak later.

I assume since Icarus is an opensource project with a public repository we have unlimited CI resources, but it's also best to be a good citizen and not waste what has been provided.

caryr (changes requested):
The reason to use two rmdir commands versus a single rm -rf is the former only removes directories making sure we have properly cleaned up the files before removing the generated directories.

caryr (changes requested):
We normally run check-vpi last.  The rational is that you only have to scroll past the VPI tests and then the Python vvp test to see the Perl vvp test summary. Once the Python and VPI tests generate a log file we can use that to create a fail summary at the end and order is not important. Until then I would like to follow the largest to smallest number of tests convention.

caryr (commented):
Should the **-Wl,-rpath=`pwd`** comment be moved to where it is now actually added?

caryr:
Could you also comment on the new build directory regarding how you are using it. I want to make sure I'm not missing anything.

I also want the just suffix and just library builds excluded to make this only two full test runs. One without a suffix or the library build and one with both.

caryr:
Yesterday I requested three changes can you take care of these so I can merge this request?

rhabacker:
> The reason to use two rmdir commands versus a single rm -rf is the former only removes directories making sure we have properly cleaned up the files before removing the generated directories.

There was a case, where the commands fail, but I do not remember the exact reason, so I will remove it.

caryr:
It can fail if things are not removed properly in a directory before it is removed. There are also odd corner cases that require manual intervention (e.g. killing a build during compilation can leave a .d file in the main directory. We may want to add that to the file removal section, but that's not required.

rhabacker:
I'm going to remove commit https://github.com/steveicarus/iverilog/pull/1310/changes/59ab77ea870f920253f6330f72998662eb2e4982 from this merge request as it turned out with #1344, that it is not possible to use a custom non root install prefix at now.

rhabacker:
> Should the **-Wl,-rpath=`pwd`** comment be moved to where it is now actually added?

As this change were added related to https://github.com/steveicarus/iverilog/commit/59ab77ea870f920253f6330f72998662eb2e4982 I removed that.

rhabacker:
> Could you also comment on the new build directory regarding how you are using it. I want to make sure I'm not missing anything.

If you are refering to https://github.com/steveicarus/iverilog/commit/59ab77ea870f920253f6330f72998662eb2e4982, it has been removed from this merge request as it turned out that it will not work until #1344 has been fixed.

> I also want the just suffix and just library builds excluded to make this only two full test runs. One without a suffix or the library build and one with both.

Which is the prefered platform to add ? 

| Distro | variant | suffix | libvvp | libveriuser| 
|--------|--------|--------|--------|--------|
| MacOSX |          | yes | yes | yes |
| Ubuntu |22.04 | no |no | yes |
| Ubuntu |24.04 | no |no | yes |
| CLANG64 | | no |no| no |
| MINGW64 | | no |no| yes |
| UCRT64 | |no | no | yes |

caryr:
libveriuser is orthogonal. Lets keep it as is. No suffix, no libvvp is one grouping and with suffix and with libvvp is the other grouping. I expect suffix and libvvp to also be orthogonal so testing them together seems like a good trade off to conserve CI time.

rhabacker:
> Maybe maybe making a 1.13.0 for stable and 1.14.0 for our current development is the best way to go.

For 


> Maybe maybe making a 1.13.0 for stable and 1.14.0 for our current development is the best way to go.

This version style has been choosen for the shared library version.

rhabacker:
> Actually one problem with making it 1.<ivl_version> is you cannot have V13 and V14 installed at the same time. Which may not be an issue for general users, but does have an impact on development testing. The advantage for general users is they can upgrade the the Icarus version without needing to recompile other tools (like NGSPICE).

This issue can be completly solved by using the `--enable-suffix` configure option. 

With `configure --enable-suffix=13` the installation looks as 
```
/home/user/src/iverilog/ci-build-suffix-native/install
/home/user/src/iverilog/ci-build-suffix-native/install/include
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/sv_vpi_user.h
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/acc_user.h
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/ivl_target.h
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/_pli_types.h
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/veriuser.h
/home/user/src/iverilog/ci-build-suffix-native/install/include/iverilog13/vpi_user.h
/home/user/src/iverilog/ci-build-suffix-native/install/lib
/home/user/src/iverilog/ci-build-suffix-native/install/lib/libvpi13.a
/home/user/src/iverilog/ci-build-suffix-native/install/lib/pkgconfig
/home/user/src/iverilog/ci-build-suffix-native/install/lib/pkgconfig/libvvp13.pc
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vvp.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/include
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/include/disciplines.vams
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/include/constants.vams
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdl_textio.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/va_math.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/ivl
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/null.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/stub.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/sizer-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/pcb-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/stub-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/ivlpp
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vvp-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vpi_debug.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/pcb.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/null-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/sizer.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdl_sys.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/blif.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/sizer.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/stub.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdlpp
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/blif-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vvp.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdl.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/v2009.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/null.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdl.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vhdl-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vlog95.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/system.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/pcb.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vlog95.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/vlog95-s.conf
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/v2005_math.vpi
/home/user/src/iverilog/ci-build-suffix-native/install/lib/ivl13/blif.tgt
/home/user/src/iverilog/ci-build-suffix-native/install/share
/home/user/src/iverilog/ci-build-suffix-native/install/share/doc
/home/user/src/iverilog/ci-build-suffix-native/install/share/man
/home/user/src/iverilog/ci-build-suffix-native/install/share/man/man1
/home/user/src/iverilog/ci-build-suffix-native/install/share/man/man1/vvp13.1
/home/user/src/iverilog/ci-build-suffix-native/install/share/man/man1/iverilog13.1
/home/user/src/iverilog/ci-build-suffix-native/install/share/man/man1/iverilog-vpi13.1
/home/user/src/iverilog/ci-build-suffix-native/install/bin
/home/user/src/iverilog/ci-build-suffix-native/install/bin/iverilog-vpi13
/home/user/src/iverilog/ci-build-suffix-native/install/bin/vvp13
/home/user/src/iverilog/ci-build-suffix-native/install/bin/iverilog13
```
and is completly independent from an install using `14` as suffix.

rhabacker:
Update
- libvvp.pc.in
  - in cflags tag add @install_suffix@ to include directory definition
  - in version tag use machine readable version setup as `major.minor`

PR #1311, opened 2026-03-23
Fix undefined behavior in schedule.cc operator delete

boomanaiden154:
Schedule.cc currently makes use of a custom slab allocator by having some structs specify a custom operator new/operator delete that call into the slab allocator. However, this setup currently relies on C++ UB, namely that writes that happen in operator delete are persisted afterwards. The slab allocator inside the free_slab function uses the memory of the object being freed to store allocator metadata, which is not allowed given the rules around operator delete.

This patch changes the internal storage of slab_allocator to a struct rather than a union so we can only return the actual storage when allocating an object and there is a header for each object that the allocator can use for metadata without writes to it being UB.

### DIFF
--- a/vvp/slab.h
+++ b/vvp/slab.h
@@ -25,8 +25,8 @@
 
 template <size_t SLAB_SIZE, size_t CHUNK_COUNT> class slab_t {
 
-      union item_cell_u {
-	    item_cell_u*next;
+      struct item_cell_s {
+	    item_cell_s*next;
 	    char space[SLAB_SIZE];
       };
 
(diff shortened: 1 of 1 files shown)


boomanaiden154:
This is not actually undefined behavior, I just can't understand the C++ standard properly.

Sorry for the noise.

caryr:
Are you withdrawing this pull request? I have not had time to review this in detail, but with the quick look I had it seemed like there would be a fix of some kind that avoided switching from the union to the struct.

boomanaiden154:
> Are you withdrawing this pull request? I have not had time to review this in detail, but with the quick look I had it seemed like there would be a fix of some kind that avoided switching from the union to the struct.

Yes. The issue I thought this fixed was not actually an issue.

I tried a method for not switching to the struct, but it seemed to result in worse devirtualization/inlining, so led to significantly worse performance on the couple of examples I tested on.

caryr:
Ok, thanks for confirming.

FYI I semi regularly test the entire test suite using valgrind to look for bad code/memory leaks. One of the machines I test with is currently using clang 19.1.7 and valgrind 3.26.0 (GIT).

PR #1315, opened 2026-03-27
Fix parallel build race with dep/ directory

### LINKED ISSUE #1314: Parallel build fails: mv to dep/ directory races with directory creation
When building with parallel make (especially with `make --shuffle`), the build fails intermittently with errors like:

```
mv: cannot move 'cprop.d' to 'dep/cprop.d': No such file or directory
mv: cannot move 'PNamedItem.d' to 'dep/PNamedItem.d': No such file or directory
mv: cannot move 'debug.d' to 'dep/debug.d': No such file or directory
```

The pattern rules in `Makefile.in`, `vhdlpp/Makefile.in`, and `vvp/Makefile.in` run `mv $*.d dep/$*.d` after compilation, but none of them list `dep` as a prerequisite. The `dep:` target exists (it just runs `mkdir dep`), but since no rule depends on it, make can start compiling before the directory is created.

The fix is to add `dep` as an order-only prerequisite (`| dep`) to each affected pattern rule. This ensures the directory exists before any compilation rule tries to move files into it, without forcing unnecessary rebuilds.

Affected files:
- `Makefile.in` : `%.o: %.cc config.h`
- `vhdlpp/Makefile.in` : `%.o: $(srcdir)/../libmisc/%.cc` and `%.o: %.cc vhdlpp_config.h`
- `vvp/Makefile.in` : `%.o: %.cc config.h` and `%.o: %.c config.h`

Related gentoo bug reports:
- https://bugs.gentoo.org/880921
- https://bugs.gentoo.org/911647
- https://bugs.gentoo.org/917344

vowstar:
Pattern rules that move .d files into dep/ do not depend on the dep directory target, so parallel make can attempt the move before the directory exists.

Add dep as an order-only prerequisite to all affected pattern rules.

Bug: https://bugs.gentoo.org/880921
Bug: https://bugs.gentoo.org/911647
Bug: https://bugs.gentoo.org/917344
Closes: https://github.com/steveicarus/iverilog/issues/1314

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -252,7 +252,7 @@ endif
 version.exe: $(srcdir)/version.c $(srcdir)/version_base.h version_tag.h
 	$(BUILDCC) $(CPPFLAGS) $(CFLAGS) $(LDFLAGS) -o version.exe -I. -I$(srcdir) $(srcdir)/version.c
 
-%.o: %.cc config.h
+%.o: %.cc config.h | dep
 	$(CXX) $(CPPFLAGS) $(CXXFLAGS) @DEPENDENCY_FLAG@ -c $< -o $*.o
 	mv $*.d dep/$*.d
 
--- a/driver/main.c
+++ b/driver/main.c
@@ -69,6 +69,9 @@ const char HELP[] =
 # include  <libiberty.h>
 #endif
 #endif
+#ifdef __APPLE__
+# include  <mach-o/dyld.h>
+#endif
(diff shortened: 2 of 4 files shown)


caryr:
This and my changes have been applied to the development (v14) branch. I plan to pull this to the V13 branch so this will be in the V13.1 release whenever that happens.

vowstar:
Thanks!

PR #1320, opened 2026-04-03
Fix spacing in compile command comment

krishnapandey08999-ui:
readme error

### DIFF
--- a/examples/hello.vl
+++ b/examples/hello.vl
@@ -25,7 +25,7 @@
   *
   *  Compile this program with the command:
   *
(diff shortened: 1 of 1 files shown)


martinwhitaker:
That isn't an error.. You can put white space between `-o` and the filename, but it isn't necessary. The `iverilog` man page shows both ways in its examples section.

caryr:
Is there a system where you have found the space was required?

caryr:
No reply so closing this.

PR #1328, opened 2026-04-27
Split aclocal macros into m4 files for aclocal-managed autoconf setup

rhabacker:
An m4 file is added in #1310. This PR lays the groundwork for managing macros using aclocal.

Used in #1310


caryr:
I assume this is just a refactor of the checking routines into separate files.

I did notice the other day that with the latest autoconf (2.73) it now defaults to c++23 instead of c++11 so we'll need to add a routine to go back to c++11.

rhabacker:
> I assume this is just a refactor of the checking routines into separate files.

Yes, previously additional macros had to be inserted into the “aclocal.m4” file, whereas now they are stored separately, which makes them easier to add and maintain.

> I did notice the other day that with the latest autoconf (2.73) it now defaults to c++23 instead of c++11 so we'll need to add a routine to go back to c++11.

Good point. I added explicit enforcement of C++11 in configure.ac (likely via AX_CXX_COMPILE_STDCXX) with  https://github.com/steveicarus/iverilog/commit/53aa471553889ad53027b36d25a620a1accd1ee5, so newer Autoconf/toolchains don’t select a newer standard implicitly.

PR #1335, opened 2026-05-02
 iverilog-vpi: Consolidate creation in driver-vpi

rhabacker:
This change standardizes the creation of iverilog-vpi-related targets, which now follow the same pattern as the iverilog targets in the 'driver' subdirectory.


caryr:
Are you planning to make the same changes for vvp?

rhabacker:
> Are you planning to make the same changes for vvp?

The build rules for the vvp binary could also benefit from a cleanup or revision, as has already been started at https://github.com/steveicarus/iverilog/pull/1310/changes/a20af5ae844455d13dcfec13a65497409f56009f. I plan to submit another merge request for this.

caryr:
We should also move the manual page to the driver-vpi directory and the rule to build the combined manual page likely needs to be fixed to handle the new location.

caryr:
FYI I just noticed the manual page was moved, it was just it had not been removed from the build directory. There were a couple issues in the top level make file. It was still trying to remove iverilog-vpi for the clean target and somehow the versioned PDF that has all the manual pages was renamed. I've pushed fixes for these so everything is good now.

PR #1338, opened 2026-05-04
Support assignment of string literals to byte arrays

larsclausen:
SystemVerilog defines a special case that allows to assign string literals to byte arrays. Each character of the string is copied to 1 element of the byte array. This is defined in section "5.9 String Literals".

The size of string literal and the byte array does not have to match. If the string literal is longer it is truncated. If it is shorter it will be padded with null-bytes.

The assignment is done left aligned, the first character ends up in the left most entry of the array. This means the order will differ whether the array is declared with ascending or descending element order.


caryr:
I think fail 4 and 5 should use bit not logic since you stated earlier 4-state is not allowed.

caryr:
FYI the commercial tool I have only allows `byte` even `bit [7:0]` fails.

caryr:
For fail 1, 4 and 5 it may be best to check both `logic` and `bit` of the given size.

larsclausen:
> FYI the commercial tool I have only allows `byte` even `bit [7:0]` fails.

We could change that, be extra strict. Some other tools like verilator seem to be even more permissive and also allow `logic [7:0]` and `bit unsigned [7:0]`.

caryr:
I was afraid of that. To me if you allow `bit [7:0]` there's no reason not to also allow `logic [7:0]` and their unsigned and possibly signed counterparts. I did a quick look at the standard and could not find where this is allowed, though I have a memory of it being allowed. Can you make sure to put a reference in this PR to where this is defined in the standard.

larsclausen:
> I was afraid of that. To me if you allow `bit [7:0]` there's no reason not to also allow `logic [7:0]` and their unsigned and possibly signed counterparts. I did a quick look at the standard and could not find where this is allowed, though I have a memory of it being allowed. Can you make sure to put a reference in this PR to where this is defined in the standard.

Its in 5.9 string literals

> A string literal can be assigned to an unpacked array of bytes. If the size differs, it is left justified.
byte c3 [0:12] = "hello world\n" ;

larsclausen:
> I was afraid of that. To me if you allow `bit [7:0]` there's no reason not to also allow `logic [7:0]` and their unsigned and possibly signed counterparts. I did a quick look at the standard and could not find where this is allowed, though I have a memory of it being allowed. Can you make sure to put a reference in this PR to where this is defined in the standard.

Sorry, I meant `bit signed [7:0]` and the reason why I went that route because according to the LRM `type(bit signed [7:0]) == type(byte)` while if you change that to logic or unsigned its not longer true. So from the LRM's point of view both types are the same type.

larsclausen:
btw. can you check what your commercial simulator does for `bit signed [7:0]`?

caryr:
Given: `bit [7:0] d = "\n" ;` is valid (1800-2023 section 5.9 top of page 82) I would assume the same should apply regarding an unpacked array.

Both `bit [7:0]` and `bit signed [7:0]` fail for the commercial tool.

larsclausen:
> Given: `bit [7:0] d = "\n" ;` is valid (1800-2023 section 5.9 top of page 82) I would assume the same should apply regarding an unpacked array.

I don't know. My strict reading of those two sentences in the LRM is that this should only work for types T where `type(T) == type(byte)`

caryr:
Yes, it seems very clear this must be a byte. I'll merge it.

PR #1341, opened 2026-05-07
vvp: Use word wide bitwise logical ops 

larsclausen:
vvp_vector4_t already defines vector wide `&=` and `|=` operators. Also add a implementation for xor (`^=`). These ops are implemented as word wide operations.

And then use those operators in vvp, rather performing bit-wise operations bit by bit.

This results in a performance improvement for these ops.

### DIFF
--- a/vvp/logic.cc
+++ b/vvp/logic.cc
@@ -94,21 +94,11 @@ void vvp_fun_and::run_run()
 
       vvp_vector4_t result (input_[0]);
 
-      for (unsigned idx = 0 ;  idx < result.size() ;  idx += 1) {
-	    vvp_bit4_t bitbit = result.value(idx);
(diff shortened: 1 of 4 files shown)


caryr:
Do we need optimized versions for nand, nor and xnor? That would avoid the second pass needed to invert the value. Also the invert is only needed for the `a` part of the number. I'll accept this as is since it does improve the performance. This basic method is behind the vector rework I started years ago, but never finished. The original Verilog expressions and 4-state layout was likely optimized to allow simple word wide operations to improve performance.

larsclausen:
The word wide `invert()` is fairly cheap. I don't think we necessarily need extra nand, nor and xnor. It might squeeze out a tiny bit more performance.

caryr:
Sounds good. We can save that optimization for if/when I get time for the vector rework.

PR #1342, opened 2026-05-07
vvp: Improve reduction operator performance

larsclausen:
The vvp reduction operators evaluate their input bit by bit. This is expensive for wide vectors.

Add word wide reduction helpers to `vvp_vector4_t` and use them for both reduction functors and vthread reduction opcodes.


caryr:
Two issues:

1. Please update the Copyright year to the current year for any files you change. I forgot to mention that on the previous PR.
2. It looks like you could us `0UL` in a couple places to make the routines more consistent.

caryr (changes requested):
See my previous comment.

larsclausen:
Fixed

PR #1354, opened 2026-05-12
Let iverilog -tvvp generates relocatable VPI module paths

### LINKED ISSUE #1344: iverilog `-tvvp` generates non-relocatable absolute VPI module paths
## Description

The `vvp` target generated by `iverilog -tvvp` embeds absolute paths for built-in VPI modules into the generated `.vvp` file.

The VVP target format documentation explicitly shows generated `:vpi_module` entries:
[[VVP target documentation](https://steveicarus.github.io/iverilog/targets/tgt-vvp.html)]

Example from the documentation:

```text id="h14n1y"
:vpi_module "/usr/local/lib/ivl/system.vpi";
```

Observed example on Windows/MSYS2:

```text id="hj8ny2"
:vpi_module "Z:\usr\X86_~FWL\sys-root\mingw\lib\ivl\system.vpi";
```

This causes generated `.vvp` files to become non-portable and non-relocatable.

## Problem

The generated paths depend on:

* the configure-time `--prefix`
* the build machine layout
* MinGW/MSYS path translation
* short-path expansion on Windows
* the installation root used during packaging

This becomes especially problematic when using staged installs via `DESTDIR`.

Example:

```bash id="ybgv6w"
./configure --prefix=/usr
make
make DESTDIR=/tmp/pkg install
```

The generated `.vvp` files still contain paths derived from the configured prefix or build environment, even if the final installation location differs.

As a result:

* `.vvp` files are not relocatable
* packaged artifacts may fail on target systems
* CI-generated outputs are machine-dependent
* reproducible builds are harder
* Windows builds leak temporary or synthetic paths

## Expected Behavior

The generated `.vvp` file should not embed absolute install paths for built-in VPI modules.

rhabacker:
Fixes #1344

Depends on #1371


caryr:
It looks like the stuff I pushed last night is failing for you even though it went in cleanly for me.. Of course I would have expected the top build directory to be `..` and everything should have just worked for UNIX like systems even if the path had embedded separators. If this doesn't fail locally try running `make clean` or maybe `make distclean`, but save the configure result from the top of `config.log` minus any of the rebuild parts at the end if they have been appended if you run `distclean` so you can rebuild everything using the same configure command.

It also looks like the verbose flag no longer prints the version and copyright information.

caryr:
Are you planning to get back to this draft?

rhabacker:
Yes, after August 8

rhabacker:
The reason is that I want to review the changes using ChatGPT Codex, and my "Free" budget will have been used up by then. If an AI for reviewing merge requests can be enabled for the iverilog project, I could continue working on that beforehand as well.

PR #1360, opened 2026-05-15
tgt-vvp: Use `%cmp/e` instead of `%cmp/u` for `case` comparisons

larsclausen:
`%cmp/e` and `%cmp/u` are very similar with `%cmp/e` not setting the lt flag and being a bit faster due to it. For case comparisons the flag is not needed so switch to `%cmp/e`. This speeds up simulation time designs which make use of case comparisons.


caryr (changes requested):
Please update the Copyright year of the file (in the future).

PR #1361, opened 2026-05-16
Reject unpacked l-value concatenation operands

larsclausen:
L-value concatenation operands must be packed values. Using an unpacked
array as an operand currently reaches the vvp assignment code and triggers
an assert.

Report an elaboration error instead.


larsclausen:
This needs a change to also cover continuous assignment

larsclausen:
> This needs a change to also cover continuous assignment

Fixed now

PR #1371, opened 2026-05-18
iverilog: add command line option -BI and -Bt

### LINKED ISSUE #1370: iverilog: missing option to run from build dir
When using `iverilog` from a build directory, it turned out that configuration files (e.g., vvp.conf) can only be found in the previously specified installation directory, and that the internal call to `ivl` requires a specific directory structure (see, for example, #1354). 

To be able to use the generated application from a build directory without prior installation, command-line options must be available to override the default settings for the aforementioned points.

rhabacker:
This PR added the command-line options mentioned above and updated the documentation for them and for the previously missing `-B<selector>` options.

In addition, a test was added to verify the new options in the `iverilog` program.

Fix #1370

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,6 +98,22 @@ AC_SUBST(EXEEXT)
 # Combined check for Microsoft-related bogosities; sets WIN32 if found
 AX_WIN32
 
+# detect msys shell
+AC_MSG_CHECKING([whether the shell is an MSYS2 runtime shell])
+
+AS_CASE([$host_os:$MSYSTEM],
+  [*mingw*:UCRT64|*mingw*:MINGW64|*mingw*:MINGW32|*mingw*:CLANG64], [
+    MSYS_SHELL=true
+    msys_shell_result=yes
+  ],
+  [
+    MSYS_SHELL=false
+    msys_shell_result=no
+  ]
+)
+AC_MSG_RESULT([$msys_shell_result])
+AC_SUBST([MSYS_SHELL])
+
 # Check to see if we are using the Sun compiler. If so then configure
 # some of the flags to match the Sun compiler syntax. This is also used
 # in the aclocal.m4 file to configure the flags used to build and link
--- a/driver/Makefile.in
+++ b/driver/Makefile.in
@@ -17,12 +17,18 @@
 #
 SHELL = /bin/sh
 
+EXEEXT = @EXEEXT@
+ENV_VVP=@ENV_VVP@
+MSYS_SHELL=@MSYS_SHELL@
+
 suffix = @install_suffix@
 
 prefix = @prefix@
 exec_prefix = @exec_prefix@
 srcdir = @srcdir@
 datarootdir = @datarootdir@
+builddir=@builddir@
+top_builddir=@top_builddir@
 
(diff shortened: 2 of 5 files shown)


caryr:
I'm not against adding the options, but does it make more sense to add a new config file to pass all this information to the various programs and then just a single option to use the results from the config file instead of the ones that are compiled into the programs? I would assume this is an all or nothing scenario so an override file may make more sense or does two sets of internal paths make more sense? One for pre-install checking and the other for post-install

martinwhitaker:
> I'm not against adding the options, but does it make more sense to add a new config file to pass all this information to the various programs and then just a single option to use the results from the config file instead of the ones that are compiled into the programs?

Like the check.conf file and the `-tcheck` option currently used to support `make check`?

caryr:
That comment was more conceptual than contextual. I've not had time to review the details of this pull request, but it seemed like we are adding a bunch of options to get checking to work before being installed. Which is good and something that repo integrators have requested previously.

I spent most of my free time this past weekend reviewing and getting other pull requests integrated along with a few post fixes, so did not have time to look at this in detail so far. The test suite supports a number of different -t options so I don't think we can just pick one. I'll post more when I have had time to review this further.

caryr:
Is there a reason the `-Bt` option is not `-BT` so it matches the other options that only take an uppercase value?

Where is the `test.conf` file created that was added to the clean target?

Do you understand why MSYS is failing?

rhabacker:
> Is there a reason the `-Bt` option is not `-BT` so it matches the other options that only take an uppercase value?

There is not really a fully consistent naming convention among the existing -B* options.

Some options use the uppercase letter because it matches the referenced subsystem name:


-BM ↔ -mmodule
-BP ↔ ivlpp
-BI ↔ ivl

Others are already semantic rather than strictly mnemonic:

-BV ↔ vhdlpp  

For -Bt, the lowercase t was chosen intentionally because it directly maps to -t <target>, which is what the option affects.

Using -BT would also be potentially confusing because Icarus Verilog already has the -T[min|typ|max] option family, so visually -BT looks very close to an extension of the existing timing-related -T* options.

So the current spelling was mainly chosen for readability and to preserve the association with -t target.

That said, if there is a preferred project-wide convention for new -B* options, I can adjust it for consistency.

rhabacker:
> Where is the test.conf file created that was added to the clean target?

That seems to be a leftover from previous tests; I'll remove it.

rhabacker:
> Do you understand why MSYS is failing?

Yes. The failure is caused by the test being executed from an MSYS runtime shell while the tested executable itself is a native Win32 program.

The MSYS shell uses POSIX-style paths such as /c/..., but the executable internally uses the native Windows API and therefore expects Windows-style paths. As a result, the test passes MSYS-style paths to a program that does not understand them.

The exclusion was therefore added because the failure is caused by the runtime environment/path format mismatch, not by the functionality being tested itself.

caryr:
> The exclusion was therefore added because the failure is caused by the runtime environment/path format mismatch, not by the functionality being tested itself.

Okay, I seem to remember there is a workaround for this and I can check it locally.

Issue #734, opened 2022-06-21
sorry: constant selects in always_* processes are not currently supported


bjourne:
I'm getting this error from this code with iverilog 11.0:
```
    always_comb begin
        if (ac >= {1'b0,y1}) begin
            ac_next = ac - y1;
            {ac_next, q1_next} = {ac_next[WIDTH-1:0], q1, 1'b1};
        end else begin
            {ac_next, q1_next} = {ac, q1} << 1;
        end
    end
```
Is the sorry message a warning or an error and does it change how the code works? As an side, it would be really useful to have a table of what System Verilog features are supported in what version and also what plans you have on adding missing features. Icarus Verilog is still a  great tool though.

martinwhitaker:
The full message is
````
sorry: constant selects in always_* processes are not currently supported (all bits will be included).
````
As far as I can see, that means the always_comb process will be triggered by changes in any bits of `ac_next`, not just `ac_next[WIDTH-1:0]`. Only you can tell if that matters in your design, but I would be surprised if it did.

On the subject of documenting what is supported, see issue #206.  Speaking just for myself, I only use standard Verilog, so have no plans to add missing System Verilog features.

**Comment by bjourne** [comment]
Thanks. It is very good to know that you are not targeting the full System Verilog spec.

**Comment by caryr** [comment]
Note that Martin was speaking for himself. Some of us have significant interest in adding support for various 
SystemVerilog constructs, but for the last number of years I have been completely consumed with designing actual circuits that many reading this comment likely use on a daily basis.

One way to remove this specific limitation is to make the constant select part of a separate assignment to a dedicated variable instead of it being inside the always construct. Another benefit of doing this is it gives the selection a name that could be more descriptive than using the select.

**Comment by steveicarus** [comment]
@bjourne Can you attach to this issue a complete Verilog source file that replicates this issue. Your snippet is fine for expressing what the issue is, but we need the code around it (such as declarations) that make this example runnable.

**Comment by chykon** [comment]
Here is another example to reproduce this issue:

```systemverilog
module example (
  input logic [7:0] d,
  output logic [3:0] q
);
  always_comb begin
    q = d[3:0];
  end
endmodule

module testbench;
  logic [7:0] d;
  logic [3:0] q;

  example example_instance (.d(d), .q(q));

  initial begin
    d = 255;
    #0;
    $display(q);
  end
endmodule
```

Askartos:
Hi
[...]

PR #1374, opened 2026-05-25
Fix #1317: Variable written in an always_ff should not driven by other process

### LINKED ISSUE #1317: [BUG] Variable written in an always_ff block must not be written by any other process
### Version
V13_0 commit dfeee909ed9f20b4870dd93423156c0170c0e1ff

### Description
Iverilog accepts an `always_ff` variable that is also assigned in an `initial` block, even though SystemVerilog requires that a variable written in an `always_ff` block must not be written by any other process.

Example:
```
    initial a = 1'b0;

    always_ff @(posedge a) begin
        a <= 1'b1;
    end
```
Per IEEE 1800 `always_ff` semantics, variables assigned inside an `always_ff` procedure shall not be assigned by any other process. An `initial` block is a separate process, so this should be rejected. However, Iverilog accepts this construct and only emits:
    Variable 'a' driven in an always_ff block, may not be driven by any other process.
This appears to be a warning rather than a hard error, allowing invalid code to compile.

### Expected behavior
Iverilog should reject the code because `a` is assigned in two separate processes:
- `initial a = 1'b0;`
- `always_ff @(posedge a) a <= 1'b1;`

A variable written in `always_ff` must not be written by any other process, including `initial`, `always`, `always_comb`, or another `always_ff`.

At minimum, this should be treated as a compile error rather than a non-fatal warning.

### Observed behavior
Iverilog accepts the design and emits only:
    Variable 'a' driven in an `always_ff` block, may not be driven by any other process.

This allows compilation of code that violates `always_ff` restrictions.

### Cross-check
- Icarus Verilog: accepts the code with no warning
- ModelSim: reject

### Reproduce
dut
```
module test;
    logic a;

    initial a = 1'b0;

    always_ff @(posedge a) begin
        a <= 1'b1;
    end
endmodule
```
iverilog command
`iverilog -g2012 -t null test.sv`

MrCookieeeee:
This PR fix #1371 

This PR adds a check for variables written by `always_ff` processes, following the SystemVerilog single-writer rule for `always_ff`.

### Implementation

1. First, it scans all `always_ff` processes and collects the `Nexus` objects written by each process through `nex_output()`. A map is used to record which `always_ff` process first writes each `Nexus`. If another `always_ff` process writes the same `Nexus`, an error is reported.

2. Second, it scans the remaining non-`always_ff` processes and checks whether any of them also write a `Nexus` that was already written by an `always_ff` process. If so, an error is reported. Compiler-scheduled initialization processes marked with `_ivl_schedule_init` are skipped, so declaration initializers are not treated as conflicting procedural writers.

### Overhead

For time complexity, this check walks the process list twice and calls `nex_output()` once for each relevant process. The map/set lookups are logarithmic, so the cost is roughly proportional to the total number of procedural write targets, with `log(N)` overhead from the containers. I think this should be acceptable for normal elaboration-time checking, but I would appreciate feedback if there is a better or more idiomatic way to do this in Icarus.

### Discussion

One limitation is that this check currently works at the `Nexus` level and does not explicitly distinguish disjoint bit/part-select writes. For example:

```systemverilog
always_ff @(posedge clk)
    q[3:0] <= a;

always_ff @(posedge clk)
    q[7:4] <= b;
```

I did not add special handling for this case yet. Please let me know whether this is necessary. If so, I could add some check on it. :D

### DIFF
--- a/elaborate.cc
+++ b/elaborate.cc
@@ -35,6 +35,8 @@
 # include  <iostream>
 # include  <sstream>
 # include  <list>
+# include  <map>
+# include  <set>
 # include  "pform.h"
 # include  "PClass.h"
 # include  "PEvent.h"
@@ -7672,6 +7674,89 @@ static void check_ff_sensitivity(const NetProc* statement)
       }
 }
 
+bool Design::check_always_ff_lvals()
+{
+      bool result = false;
+      map<const Nexus*, const NetProcTop*> always_ff_lvals;
+
+      for (NetProcTop*pr = procs_ ; pr ; pr = pr->next_) {
+	    if (pr->type() != IVL_PR_ALWAYS_FF)
+		  continue;
+
+	    NexusSet outputs;
+	    pr->statement()->nex_output(outputs);
+
+	    set<const Nexus*> seen_outputs;
+	    for (unsigned idx = 0 ; idx < outputs.size() ; idx += 1) {
+		  const NexusSet::elem_t&output = outputs[idx];
+		  const Nexus*nexus = output.lnk.nexus();
+		  if (! seen_outputs.insert(nexus).second)
+			continue;
+
+		  auto cur = always_ff_lvals.find(nexus);
+		  if (cur == always_ff_lvals.end()) {
+			always_ff_lvals[nexus] = pr;
+			continue;
+		  }
+
+		  if (cur->second == pr)
+			continue;
+
+		  const char*name = nexus->name();
+		  cerr << cur->second->get_fileline() << ": error: "
+			  "variable '" << name << "' written in an "
+			  "always_ff process may not be written by any "
+			  "other process." << endl;
+		  cerr << pr->get_fileline() << ":      : "
+			  "process also writes variable '" << name
+		       << "'." << endl;
+		  result = true;
+		  errors += 1;
+	    }
+      }
+
+      if (always_ff_lvals.empty())
+	    return result;
+
+      for (NetProcTop*pr = procs_ ; pr ; pr = pr->next_) {
+	    if (pr->type() == IVL_PR_ALWAYS_FF)
+		  continue;
+
+	    if ((pr->type() == IVL_PR_INITIAL) &&
+	        (pr->attribute(perm_string::literal("_ivl_schedule_init"))
+	           .as_ulong() != 0))
+		  continue;
+
+	    NexusSet outputs;
+	    pr->statement()->nex_output(outputs);
+
+	    set<const Nexus*> seen_outputs;
+	    for (unsigned idx = 0 ; idx < outputs.size() ; idx += 1) {
+		  const NexusSet::elem_t&output = outputs[idx];
+		  const Nexus*nexus = output.lnk.nexus();
+		  if (! seen_outputs.insert(nexus).second)
+			continue;
+
+		  auto cur = always_ff_lvals.find(nexus);
+		  if (cur == always_ff_lvals.end())
+			continue;
+
+		  const char*name = nexus->name();
+		  cerr << cur->second->get_fileline() << ": error: "
+			  "variable '" << name << "' written in an "
+			  "always_ff process may not be written by any "
+			  "other process." << endl;
+		  cerr << pr->get_fileline() << ":      : "
+			  "process also writes variable '" << name
+		       << "'." << endl;
+		  result = true;
+		  errors += 1;
+	    }
+      }
+
+      return result;
+}
+
 /*
  * Check to see if the always_* processes only contain synthesizable
  * constructs.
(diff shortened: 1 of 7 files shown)


ivtest/regress-sv.list
```diff
@@ -81,6 +81,7 @@ always_ff_fail		CE,-g2005-sv		ivltests
 always_ff_fail2		CE,-g2005-sv		ivltests
 always_ff_fail3		CE,-g2005-sv		ivltests
 always_ff_fail4		CE,-g2005-sv		ivltests
+always_ff_lval_fail	CE,-g2005-sv		ivltests
```
larsclausen: for the new test can you use newstyle test format using json. See ivtest/regress-vvp.list and ivtest/vvp_tests

elaborate.cc
```diff
@@ -7672,6 +7674,95 @@ static void check_ff_sensitivity(const NetProc* statement)
...
+      map<const Nexus*, const NetProcTop*> always_ff_lvals;
+
+      for (NetProcTop*pr = procs_ ; pr ; pr = pr->next_) {
+	    if (pr->type() != IVL_PR_ALWAYS_FF)
+		  continue;
+
+	    NexusSet outputs;
+	    pr->statement()->nex_output(outputs);
+	    if (outputs.size() == 0)
+		  continue;
```
larsclausen: probably dont need this extra check since the loop below will just not run if outputs.size() is 0. Same for the other loop below

elaborate.cc
```diff
@@ -7672,6 +7674,95 @@ static void check_ff_sensitivity(const NetProc* statement)
...
+
+	    set<const Nexus*> seen_outputs;
+	    for (unsigned idx = 0 ; idx < outputs.size() ; idx += 1) {
+		  const NexusSet::elem_t&output = outputs[idx];
+		  const Nexus*nexus = output.lnk.nexus();
+		  if (! seen_outputs.insert(nexus).second)
+			continue;
+
+		  map<const Nexus*, const NetProcTop*>::const_iterator cur
```
larsclausen:
```suggestion
		  auto cur
```

to keep the line a bit shroter

MrCookieeeee:
Hi,

I have updated the code and resubmitted the PR.

The changes are:

1. In `elaborate.cc`, I removed the redundant `if (outputs.size() == 0)` check and changed the iterator declaration to use `auto cur` to keep the line shorter.

2. I added the new regression test `always_ff_lval_fail` to `regress-vvp.list`, together with the corresponding `always_ff_lval_fail.json` and `always_ff_lval_fail-iverilog-stderr.gold` files.

Thanks.

ivtest/regress-sv.list
larsclausen: you added the json test, this line should be remove now. we only have the test in either regress-vvp.list or regress-sv.list

elaborate.cc
larsclausen: you changed one of the iterator declarations to auto, but this one where the comment was is still the long form

MrCookieeeee:
Copy, i will fix it real quick.

larsclausen (approved):

MrCookieeeee:
Hi, do you have any plan to merge this PR?

caryr:
I've been busy, but here are some comments.

It looks like with a bit more work you can create a common parameterized call for each of the passes to avoid the significant code duplication. Adding the value and checking the initialization look to be the major differences between the two passes. 

Why are always_comb and always latch not also supported?

We really should support the longest static prefix so the actual bit being assigned is known. This could be by parameterizing the nex_output() routines to adapt to the proper evaluation for always_comb/always_ff/always_latch. Having this should also fix the bug that selects are currently missed when calculating the sensitivity. See #734. Until this is supported you cannot make this an error if a select is used. Your example in the description is perfectly valid, but I assume will generate an error

MrCookieeeee:
Copy that. Thanks for the carefull review, i will try to figure it out.
It may take couple of days. :D

caryr:
Have you been able to look at this or do you need some suggestions? I'd like to get this turned in.

MrCookieeeee:
Hi, sorry for the delay. I have 
(1) removed the duplication
> It looks like with a bit more work you can create a common parameterized call for each of the passes to avoid the significant code duplication. Adding the value and checking the initialization look to be the major differences between the two passes.


(2) and supported always_comb and always_latch
> Why are always_comb and always latch not also supported?

Above parts are easy, but adding width awareness when calculating the sensitivity, together with fixing #734 , needs loads of change. I'm still working on it.

caryr:
I don't see an updated pull request with your latest changes, am I missing something?

Also the messages are about the processes when the standard rules are really about the assignments. They are also using the last item found as the reference event when normal SystemVerilog treats things declared first as the reference and later things as overrides or duplicates. All the information is there, but it feels backwards.

The current patch still has these as hard errors. I think this will falsely trigger an error when constant bit selects are used in different always_* blocks. Since we don't properly support constant selects can we at least detect these and switch them to a warning. I think the general rule would be if everything is a bit/part select then we issue a warning, but if anything covers the entire element then it's an error. This may be too complicated to figure out so we may have to switch them all to warnings until the longest-static-prefix issue is solved.

PR #1377, opened 2026-05-28
Fix #1322: No warning on dynamic Out-Of-Bound array access

### LINKED ISSUE #1322: No warning on dynamic Out-Of-Bound array access
### Summary

Out-of-bounds writes to fixed-size unpacked arrays currently execute as silent no-ops in Icarus Verilog, which is compliant with IEEE 1800-2017 §7.4.6. However, the lack of any diagnostic makes these bugs difficult to detect, especially when designs use indices wider than the array bounds. In contrast, other tools (e.g., Verilator) emit warnings—although their execution semantics are incorrect. Icarus Verilog should retain its correct no-op behavior but emit a warning when an invalid index is used for an array write.

### Root Cause Analysis

The issue arises when an unpacked array is indexed by a variable whose value can exceed the valid index range. For example, a 4-element array indexed by a 4-bit counter will produce out-of-bounds accesses once the counter exceeds 3. Per IEEE 1800-2017 §7.4.6, such writes must be ignored (no operation). Icarus Verilog correctly implements this semantic, but does not issue any warning or diagnostic. This creates a silent failure mode where design intent is violated without visibility, complicating debugging and masking potential RTL bugs.

### Reproduction
```
module tb;

  logic clk = 0;
  always #1 clk = ~clk;

  logic [7:0] shift_reg [3:0];  // 4 elements
  logic [3:0] counter;          // 0–15 (will exceed bounds)

  initial begin
    // initialize
    for (int i = 0; i < 4; i++) shift_reg[i] = 8'hA0 + i;
    counter = 0;

    // perform 6 writes (2 are OOB)
    repeat (6) begin
      @(posedge clk);
      shift_reg[counter] <= 8'hFF;
      counter <= counter + 1;
    end

    // check contents
    #1;
    $display("shift_reg = %p", shift_reg);

    // expected (IEEE-compliant behavior: no-op on OOB)
    if (shift_reg[0] !== 8'hFF ||
        shift_reg[1] !== 8'hFF ||
        shift_reg[2] !== 8'hFF ||
        shift_reg[3] !== 8'hFF) begin
      $fatal("FAIL: array contents corrupted");
    end

    $display("PASS");
    $finish;
  end

endmodule
```

Running under Icarus Verilog:
```
iverilog -g2012 -o test.vvp oob_array_mismatch.sv
vvp test.vvp
```

Simulation completes without warnings, and out-of-bounds writes are silently dropped, result in pass.

MrCookieeeee:
As discussed in #1322, Icarus already reports out-of-bounds accesses for constant unpacked-array indices, such as:
```verilog
assign a[10] = 1;
```
However, the same diagnostic is currently missing for variable-indexed unpacked-array writes, where the index value is only known at runtime.

This PR extends `-Wselect-range` to dynamic out-of-bounds unpacked array writes in the VVP backend/runtime. Key changes are as follows:

1. Add `warn_select_range` and `array_write_warn_suffix()`,  emit `/
  w` opcode variants only when the warning is enabled.
2. Cover blocking/immediate array writes via `%store/vec4a/w`, `%store/reala/w`,
  `%store/stra/w`, and `%store/obja/w`.
  Cover nonblocking/delayed/event array writes via `%assign/ar*/w` and `%assign/vec4/a/*/w`.
3. Add some regression tests

If you have a better implementation approach, I would be happy to adjust the PR. I want to keep the change both elegant and efficient.


caryr:
I will not have time to review this and your other submission for about a week.

MrCookieeeee:
No worries, take your time.

larsclausen:
Not sure we need separate instructions for this. We either want this for all array access or not, there is no selective allow it for this, but dont allow it for that. Maybe that is better handled as a vvp warning flag. But lets see what others say

MrCookieeeee:
> Not sure we need separate instructions for this. We either want this for all array access or not, there is no selective allow it for this, but dont allow it for that. Maybe that is better handled as a vvp warning flag. But lets see what others say

I agree. Adding new instruction like assign/vec4/a/d/w also cause me lots of trouble, and also make the repo kinda repeatitive. Maybe you can just decline this PR, or i will push a new verison of it later.

tgt-vvp/vvp_process.c
```diff
@@ -54,24 +53,23 @@ static void assign_to_array_r_word(ivl_signal_t lsig, ivl_expr_t word_ix,
...
-	    fprintf(vvp_out, "    %%jmp/0 t_%u, 4;\n", do_assign);
-	    fprintf(vvp_out, "    %%pop/real 1;\n");
-	    fprintf(vvp_out, "    %%jmp t_%u;\n", end_assign);
-	    fprintf(vvp_out, "t_%u ;\n", do_assign);
       }
 
       if (dexp != 0) {
 	      /* Calculated delay... */
 	    int delay_index = allocate_word();
+	    int index_flag = allocate_flag();
```
MrCookieeeee, line 65: Explaination: Evaluating the delay expression may overlap flag 4, which currently records the array index status. Save and restore it later.

MrCookieeeee:
Hi, could anyone review this PR?

PR #1387, opened 2026-06-19
Preserve sign of negative zero

flaviens:
Hi!

The function `draw_realnum_real()` detected the sign via `(value < 0)`, which is false for IEEE 754 -0.0. Therefore, a source/constant-folded -0.0 was emitted as +0.0. Use `signbit()` to set the sign bit, matching the runtime real value and
IEEE 754-2019 sec 3.4 / IEEE 1800-2017 sec 6.12.1.

Thanks!
Flavien

### DIFF
--- a/ivtest/ivltests/real_negative_zero.v
+++ b/ivtest/ivltests/real_negative_zero.v
@@ -1,0 +1,35 @@
+// Check that the vvp code generator preserves the sign of a negative zero
+// real constant. The sign used to be detected with (value < 0), which is
+// false for IEEE 754 -0.0, so a -0.0 constant was emitted as +0.0 and the
+// compiled value no longer matched the runtime real value.
+
+module test;
+
+  real nz;
+  real pz;
+
+  initial begin
+    nz = -0.0;
+    pz = 0.0;
+
+    // The sign bit is the only thing that tells -0.0 from +0.0.
+    if ($realtobits(nz) !== 64'h8000_0000_0000_0000) begin
+      $display("FAILED: -0.0 stored as %h", $realtobits(nz));
+      $finish;
+    end
+
+    if ($realtobits(pz) !== 64'h0000_0000_0000_0000) begin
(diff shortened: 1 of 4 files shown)


caryr:
When I compared this with the code in the vlog95 generator it is using signbit() for detecting negative 0.0 and also for detecting a negative NaN. My guess is this also needs to be done for NaN in the VVP code generator. Can you check if a change is needed for NaN? I'll check later if you do not get around to it.

caryr:
Oops, the other branch in the vlog95 generator that used signbit() was for the infinities (likely not needed, but maybe faster), though I thought NaN can also be positive or negative. Now I definitely need to investigate more.

larsclausen:
code change looks, ok to me. Can you also add a regression test that checks that this case is handled correctly?

larsclausen:
> When I compared this with the code in the vlog95 generator it is using signbit() for detecting negative 0.0 and also for detecting a negative NaN. My guess is this also needs to be done for NaN in the VVP code generator. Can you check if a change is needed for NaN? I'll check later if you do not get around to it.

`$realtobits(-nan)` is something where we output the wrong value atm. For pretty much everything else the distinction between `nan` and `-nan` does not really matter.

caryr:
I agree the difference between nan and -nan is mostly immaterial. What I am looking for is consistency between the underlying C/C++ and how the result is generated in Icarus. There are places where not handling the sign bit for nan creates differences, but then again the C/C++ code is not what I would consider consistent either (the result can depend on the optimization selected during compilation). My biggest concern is that the compiler can produce different results than the runtime because we are not preserving the nan sign.  For example: 0.0/0.0 is producing -nan in the runtime, but because the sign is dropped nan in the compiler. Even more wonky is when compiled with -O2 the C code produces nan for -0.0/0.0. It also depends on the compiler used, gcc versus clang.

I agree with Lars, add a regression test and we will commit this.

PR #1388, opened 2026-06-19
tgt-flow: add a dataflow (.flow) exporter target

wm2015email:
## Summary

Adds `tgt-flow`, a code-generator target that exports the elaborated
design as a JSON **dataflow database** (`iverilog -tflow -o design.flow`)
instead of a simulation/netlist artifact. It describes the module
hierarchy, each module's ports/parameters/signals, and the dataflow
graph (which processes and continuous assignments read/drive which nets),
so IDEs and analysis tools can render dataflow panels, browse the
elaborated hierarchy, and trace input→output continuity.

The schema (`flowtracer1.verilog.v0`) is the Verilog companion to the
GHDL `--flow` exporter's `flowtracer1.vhdl.v0`, so a single consumer
works across Verilog and VHDL designs.

## Source positions

Every `pos`/`inst_pos` is a compact byte-offset string `"start:begin:end"`
(parts empty when not applicable) to help source-scanning consumers
locate constructs:

- port/signal → `"start::"`
- module → `"start::end"` (`module` .. `endmodule`)
- instance → `"start::end"` (.. `;`)
- `always`/`initial` block or generate frame → `"start:begin:end"`

Because `ivl_target` exposes only line numbers, the byte offsets and
begin/end spans are recovered by a small comment/string-aware scanner
that reads the source files (cached, with a per-file line table).

## Tests / docs

- `ivtest/flow.list` + `ivtest/flow_reg.py`: compiles each `flow/*.v`,
  parses the JSON, and asserts schema, hierarchy, dataflow facts
  (clocked processes, port-map continuity, continuous-assign cone reads,
  generate frames), and that every position is well formed. All pass.
- `Documentation/targets/tgt-flow.rst` documents the target and the
  position format.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Documentation/targets/index.rst
+++ b/Documentation/targets/index.rst
@@ -21,3 +21,4 @@ Icarus Verilog. The code generator is selected by the "-t" command line flag.
    tgt-sizer
    tgt-verilog
    tgt-blif
+   tgt-flow
--- a/Documentation/targets/tgt-flow.rst
+++ b/Documentation/targets/tgt-flow.rst
@@ -9,9 +9,10 @@ assignments read and drive which nets), so that IDEs and analysis tools
 can render dataflow panels, browse the elaborated hierarchy, and trace
 input-to-output continuity across the design.
 
-The emitted schema (``flowtracer1.merged.v0``) is shared with the GHDL
-``--flow`` exporter, so a single consumer works across Verilog (via this
-target) and VHDL (via GHDL) designs.
+The emitted schema (``flowtracer1.verilog.v0``) is the Verilog companion
+to the GHDL ``--flow`` exporter (``flowtracer1.vhdl.v0``), so a single
+consumer works across Verilog (via this target) and VHDL (via GHDL)
+designs.
 
 
 USAGE
@@ -34,7 +35,7 @@ OUTPUT
 The output is a single JSON object with these top-level fields:
 
 ``schema``
-    Always ``"flowtracer1.merged.v0"``.
+    Always ``"flowtracer1.verilog.v0"``.
 
 ``top``
     The basename of the root module instance.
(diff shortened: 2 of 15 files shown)


caryr:
A couple comments.

- It looks like you switched from a merged schema to a Verilog specific schema, but did not update the documentation. It would also be better if the documentation described a complete flow and had example diagrams, etc.
- I looked for the GHDL version of this to see if there were examples of the output or how the output is used and did not find this. Having more context would help understand where this goes.
- The flow file should be split into sections: maybe something like the following: top level, getting the elaboration data from the compiler, getting the extra data and emitting the data. I'm sure this works as is, but it's harder to review and change when it is written as one large file.
- What about generate case?
- How are multiple top level modules supported?
- Probably more, but this is all I have after a quick review.

wm2015email:
Thanks @caryr, these are all fair — appreciate the careful read. Responses point by point:

  1. Verilog-specific schema vs. docs. You're right, the docs lagged the schema change. When I split the merged schema into language-specific ones
  (flowtracer1.verilog.v0 here, flowtracer1.vhdl.v0 on the GHDL side) I didn't update tgt-flow.rst to match. I'll fix that and, as you suggest, expand it
  into a complete walkthrough: a small worked example (source → iverilog -tflow → annotated .flow JSON), a field-by-field description of the schema, and a
  diagram showing how the hierarchy + dataflow graph map onto an IDE dataflow panel.

  2. GHDL version / where this goes. Good point that there's no context for the consumer. The .flow file feeds a dataflow viewer using javescript version of ELK layout engine to place the boxes and 
  renders module hierarchy, per-module port/signal panels, and input→output continuity traces; the whole reason for the shared schema is that one viewer
  works across both Verilog (-tflow) and VHDL (GHDL --flow). I'll add to the docs: an example .flow output, a short "how it's consumed" section, and a
  cross-reference to the GHDL --flow exporter so the two halves are discoverable together. (The GHDL side is a separate PR; I'll link them.)

  3. One large file. Agreed — flow.c grew into one file that's hard to review. I'll split it along the phases you describe:
  - compiler-data extraction (walking ivl_scope/ivl_signal/processes/cont-assigns),
  - the supplemental source-span scanner (the comment/string-aware byte-offset recovery),
  - JSON emission (envelope, hierarchy, modules, dataflow),

  with the top-level target_design orchestrating them. That should make each part reviewable and changeable on its own.

  4. Generate. Generate scopes are handled, just folded rather than given first-class objects. In the hierarchy tree they're emitted as transparent frame
  nodes (with their "start:begin:end" span), and their elaborated contents — signals, continuous assigns, processes — fold into the enclosing module's
  dataflow so input→output continuity is preserved across the generate boundary. The per-module "generates": [] field is currently a placeholder; the
  generate structure lives in the hierarchy frames. flow_reg.py covers the generate-frame case. If you'd prefer generates represented as explicit per-module
  objects too, I'm happy to populate that.

  5. Multiple top-level modules. Supported. target_design calls ivl_design_roots() and loops over every root for each phase (collect → assign net ids →
  build cells → emit), so each top module is collected and emitted independently into the same .flow file. I'll add a multi-top test to flow.list to lock
  that in.

  I'll push the doc rewrite + the file split as follow-up commits and ping you when they're up.

wm2015email:
use_case_for_updates

wm2015email:
use_case2

caryr:
Sounds good. Do you handle out of module references or accessing package data? Out of module is fairly common and can create a dotted connection between modules. Packages are a bit more obscure, but they can be used to exchange data between modules without a known physical connection.

As a side comment: I've always been interested in using graphviz to generate flows like this, but have never taken the time to write the target.

caryr:
Generate creates a new scope, but so do named blocks so you should likely handle them and their local variables consistently.

caryr:
Is this ready? I asked a couple questions and no response. I would like to get this pulled into the repo if it's ready.

caryr:
Do you have answers for my previous questions? I would like to get this added to the repo.

PR #1389, opened 2026-06-20
vpi: extended VCD ($dumpports) writer + native structural VPI (process/primitive/driver/contassign)

wm2015email:
## Summary

Two related additions to Icarus's VPI, on top of one another:

1. **Extended VCD (`$dumpports`) writer** — a `system.vpi` task family
   (`$dumpports`, `$dumpportsoff/on/all/flush/limit`) that writes an IEEE
   1364-2005 Clause 18 extended VCD recording a module instance's ports with
   direction and drive strength. Output is byte-compatible with the GHDL
   `--evcd` writer, so one reader works across Verilog and VHDL.

2. **Native structural-connectivity VPI** — the IEEE 1364 object model that
   Icarus previously left unimplemented, so structural consumers (dataflow /
   waveform / analysis tools) get real data instead of empty fallbacks. Each
   object is sourced faithfully from the compiler and threaded
   `ivl` → tgt-vvp → `.vvp` directive → assembler → runtime `__vpi*` → scope
   iterator → `vpi_user.h`:
   - **`vpiProcess`** (always/initial/final): type, location, scope.
   - **`vpiPrimitive` / `vpiPrimTerm`** (gate/switch/UDP): `vpiPrimType`,
     `vpiSize`, terminals with `vpiDirection`/`vpiTermIndex`.
   - **`vpiDriver`**: per-driver value (`vpiStrengthVal`) and scope, read from
     the tri resolver (tagged at elaboration via `.resolv_drv`).
   - **`vpiContAssign`** (+`vpiLhs`/`vpiRhs`): continuous assignments preserved
     through the frontend (`NetScope` → `dll_target::end_design` → new
     `ivl_scope_cassigns`/`ivl_cassign_*` API → `.contassign` →
     `__vpiContAssign`). `vpiLhs`/size/location always faithful; `vpiRhs` is the
     r-value net for simple assigns, null for an expression r-value.

Building on `vpiDriver`, the `$dumpports` writer now separates an inout's
module-side drive from its external drive and emits the full IEEE 1364-2005
§18.4.3 conflict-state characters (`0/1/?/F`, the `A/a/B/b/C/c` conflicts, and
`d/u/l/h` by drive strength) rather than collapsing a genuine bus conflict to
`?`.

## Constants

New object/relationship/property codes are added to `vpi_user.h` using the
standard IEEE values where they are free in Icarus's numbering, and
Icarus-local values (continuing the 13x block) where the standard value
collides with an existing Icarus code (noted inline).

## Tests

`ivtest` regressions: `evcd_basic`, `evcd_bus`, `evcd_onoff`, and `evcd_inout`
(the inout conflict states) all pass. The full Verilog regression
(`regress-vlg.list`) stays clean — the remaining failures are pre-existing and
unrelated.

## Notes / follow-ups

- `vpiLoad` and full expression objects for `vpiRhs` are left as follow-ups.
- A single-driver inout (no resolver) reports the resolved-value EVCD char
  rather than the directional `H`/`U`; inouts driven from both sides (the
  normal testbench case) get the full set.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Documentation/usage/waveform_viewer.rst
+++ b/Documentation/usage/waveform_viewer.rst
@@ -149,6 +149,16 @@ output ``L``/``H``, three-state, etc.) and the strength0/strength1 components.
 The companion tasks $dumpportsall, $dumpportsoff, $dumpportson,
 $dumpportsflush, and $dumpportslimit mirror the corresponding $dumpall family.
 
-The extended VCD format is byte-compatible with the GHDL ``--evcd`` writer, so
-the same waveform reader can consume port dumps from both Verilog and VHDL
-designs.
+For ``inout`` ports, Icarus separates the module-side drive from the external
+(testbench-side) drive and emits the full IEEE 1364-2005 conflict-state
+characters: ``0``/``1`` (both sides agree), ``A``/``a``/``B``/``b``/``C``/``c``
+(the two sides drive conflicting values), ``d``/``u``/``l``/``h`` (same value
+but differing drive strength), ``F`` (neither side drives), and the
+directional ``D``/``U``/``H``/``L`` forms when only one side drives. This is
+more precise than a plain resolved value, which would collapse a genuine bus
+conflict to ``?``.
+
+The extended VCD format is otherwise byte-compatible with the GHDL ``--evcd``
+writer, so the same waveform reader can consume port dumps from both Verilog
+and VHDL designs. (GHDL does not separate inout drive sides, so it emits the
+resolved-value states only; Icarus is a strict superset there.)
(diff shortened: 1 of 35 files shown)


caryr:
How odd. I was reading the extended dump part of the standard earlier wondering if it was ever going to be needed in Icarus.

Regarding the duplicate value in vpi_user.h there should only be a value collision if they are used as part of the same property. If we have inadvertently coded something with the wrong value that should be fixed otherwise we should be following what is given in the standard. This is true even if the values are the same.

caryr:
Some test failures to look at and to get windows to compile you will need to add the new routines to the appropriate .def file. One of the oddities of windows versus Unix based systems.

caryr:
Have you been able to look at my comments above?

PR #1395, opened 2026-06-21
Support classes in generate blocks

### LINKED ISSUE #1384: Crash (failed assertion) when a class is declared inside a generate block
Declaring a class inside a generate block makes iverilog abort:

```verilog
module m;
  if (1) begin : g
    class C;
    endclass
  end
endmodule
```

```
$ iverilog -g2012 bug.sv
bug.sv:3: assert: pform.cc:588: failed assertion !pform_cur_generate
Aborted (core dumped)
```

The same class at module scope compiles fine, so it looks specific to the generate scope. Also happens with `-g2009` and `-g2005-sv`. Built from current master (14.0 devel, de415b2). I'd expect it to either accept the class or print a normal error rather than abort.

Found with DepTyCheck (https://github.com/buzden/deptycheck).

larsclausen:
SystemVerilog allows class declarations as module and generate items. Currently a class declaration in a generate block triggers an assert because `pform_push_class_scope()` only records classes in `PScopeExtra` scopes.

Add class storage to `PGenerate` and elaborate those classes like module and package classes. When registering task, function or class declarations, only use the current `PGenerate` object as the target if it is also the current lexical scope. This distinction matters for generated classes because `pform_cur_generate` remains set while the class body is parsed, but the current lexical scope has changed to the `PClass`. This records the class declaration in the generate block while leaving methods and constructors in the class scope.

Resolves #1384

### DIFF
--- a/PGenerate.h
+++ b/PGenerate.h
@@ -26,10 +26,12 @@
 # include  <list>
 # include  <map>
 # include  <valarray>
+# include  <vector>
 # include  "pform_types.h"
 
 class Design;
 class NetScope;
+class PClass;
 class PExpr;
 class PFunction;
 class PProcess;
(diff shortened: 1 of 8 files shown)


caryr:
There are merge conflicts that need to be resolved.

PR #1402, opened 2026-06-22
vvp: Add opcode for unary real minus

larsclausen:
Currently the vvp target emits unary real minus as `0.0 - value`. This is not the same operation for all real values. It loses the negative zero result for `-(+0.0)` and does not reliably flip the sign bit for NaN values whose bits are visible through `$realtobits`.

Add `%neg/wr` and use it for unary real minus. This performs a direct negation of the real stack value, so zero, NaN and infinity all use the same operation as unary minus instead of a binary subtraction from zero.

### DIFF
--- a/Documentation/developer/guide/vvp/opcodes.rst
+++ b/Documentation/developer/guide/vvp/opcodes.rst
@@ -812,6 +812,10 @@ result is pushed back on the vec4 stack.
 
 This opcode multiplies two real words together.
 
+* %neg/wr
+
+This opcode negates the real value on top of the real stack.
+
 * %nand
 
 Perform the bitwise NAND of two vec4 vectors, and push the result. Each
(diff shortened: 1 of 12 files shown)


caryr:
Do we also need fixes in the compiler and how these constants are passed to the runtime? I know -nan is not passed from the compiler to runtime properly. I assume there could be other places where there are issues.

caryr:
Looks good, but does unary minus always work? It looks like your tests are complete so if any systems fail we'll know. It's more what does the floating point standard guarantee? I've not had time to read the FP standard in detail. I'll merge and test on systems not covered by our CI.

larsclausen:
For anything IEEE-754 compliant this should work. This is in section 6.3 of the IEE-754, if I read it correctly negate(x) is supposed to flip the bit even when its a NaN. I guess C++ could chose to not follow it, i.e. when compiling with -ffast-math.

PR #1404, opened 2026-06-24
Fix assert on constant == with unequal-length string operands

sifferman:
Hello. Thanks for this tool!

I ran into this bug while using `generate if` on `parameter string`s.

Thanks!

### DIFF
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -916,6 +916,17 @@ NetExpr* PEBComp::elaborate_expr(Design*des, NetScope*scope,
       eval_expr(lp, l_width_);
       eval_expr(rp, r_width_);
 
+	// test_width() does not equalize operand widths for string
+	// operands, but constant evaluation expects equal widths.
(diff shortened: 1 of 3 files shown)


caryr:
This certainly fixes the issue, but it is missing a few things:

1. The assertion is intended to verify that the compiler has properly sized the constants so we only want to extend the operands when they are strings constants. I'm not sure how to detect this since string constants are showing up as `bool`
2. This does not fix the case equality or wildcard equality operators.

I need to think if there is a more general solution for this. I'm also a bit surprised this is not part of the base `verinum` functionality.

sifferman:
Thanks, @caryr! I now moved the fix to `elaborate_expr()` in elab_expr.cc.

I couldn't put it in `test_width()` because it runs before elaboration, before a string is known to be constant or a signal. So the comparison would have to always be done in runtime.

caryr:
`elaborate_expr()` is likely where this belongs, so this is fine.

PR #1405, opened 2026-06-24
Add missing AH_TEMPLATE entries for autoheader compatibility

wotomchen:
The AC_DEFINE macros for HAVE_LROUND, HAVE_LLROUND, HAVE_NAN,
HAVE_FMIN, HAVE_FMAX, and UINT64_T_AND_ULONG_SAME lack AH_TEMPLATE
declarations, causing autoheader to fail with newer autoconf versions.
Add the corresponding AH_TEMPLATE calls so autoheader can generate
proper config.h.in templates.

Also add the missing #undef entries to the hand-maintained config.h.in
to match.

### DIFF
--- a/config.h.in
+++ b/config.h.in
@@ -35,8 +35,13 @@
 # undef HAVE_LIBZ
 # undef HAVE_LIBBZ2
 # undef HAVE_LROUND
+# undef HAVE_LLROUND
+# undef HAVE_NAN
+# undef HAVE_FMIN
+# undef HAVE_FMAX
 # undef HAVE_SYS_WAIT_H
 # undef WORDS_BIGENDIAN
+# undef UINT64_T_AND_ULONG_SAME
 
 #ifdef HAVE_INTTYPES_H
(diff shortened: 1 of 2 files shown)


caryr:
The changes to configure.ac seem reasonable, but does config.h need to be changed?

There are a bunch of config files and individual ones use various defines as needed. In aggregate all the defines should be used across the various files, but they are not currently used by all config files. Maybe there is a better way.

I believe the following is the complete list of configuration files created by configure:
```
./_pli_types.h.in
./version_base.h.in
./config.h.in
./tgt-vhdl/vhdl_config.h.in
./vhdlpp/vhdlpp_config.h.in
./vpi/vpi_config.h.in
./tgt-pcb/pcb_config.h.in
./vvp/config.h.in
./libveriuser/config.h.in
./tgt-vvp/vvp_config.h.in
```

caryr:
Any answer to my previous question? I'd like to get this added.

caryr:
Any answer to my question?

PR #1406, opened 2026-06-24
musl/OHOS compatibility: dlopen detection, _LARGEFILE64_SOURCE, config.guess

wotomchen:
Three changes for musl libc and OpenHarmony (OHOS) compatibility:

1. dlopen detection - Check for dlopen in libc before trying -ldl. On musl (and OHOS), dlopen is part of libc with no separate libdl.

2. _LARGEFILE64_SOURCE fallback - On musl, fopen64 needs _LARGEFILE64_SOURCE to be declared; glibc exposes it via _LARGEFILE_SOURCE alone. Add a fallback check so fopen64 is found on musl.

3. config.guess OHOS detection - Recognize HarmonyOS from uname -s and canonicalize as *-unknown-linux-ohos.

### DIFF
--- a/config.guess
+++ b/config.guess
@@ -940,6 +940,9 @@ EOF
     *:SerenityOS:*:*)
         GUESS=$UNAME_MACHINE-pc-serenity
         ;;
+    *:HarmonyOS:*:*)
+        GUESS=$UNAME_MACHINE-unknown-linux-ohos
+        ;;
     *:Interix*:*)
 	case $UNAME_MACHINE in
 	    x86)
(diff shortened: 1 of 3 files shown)


caryr:
This pull request also includes the `AH_TEMPLATE` changes from the previous PR. For `config.guess` this change should be pushed into the master repository. I did update to the latest available, but I do not see the Harmony updates in the latest file. Can you push that part of the change upstream as described in the `config.guess` file? I can pull from upstream when that is done.

caryr:
Are you planning to push these changes to the `config.guess` master repository? I'd like to get this applied.

PR #1412, opened 2026-06-28
vvp: zero reused VPI result buffer in thread-vector value getter

### LINKED ISSUE #1411: $countones() gives wrong result when used inline in an expression (e.g. `$countones(x) == 1`)
`$countones` returns the wrong value when used directly inside an expression, like `$countones(x) == 1`. Assigning it to a variable first and then comparing works.

```systemverilog
module tb;
  logic [3:0] an = 4'b1110;   // ~an = 0001, so $countones(~an) == 1
  int c;
  initial begin
    c = $countones(~an);
    $display("via variable : %0d   (correct: 1)", c);
    $display("inline ==1   : %b   (should be 1)", ($countones(~an) == 1));
  end
endmodule
```

Output:

```
via variable : 1   (correct: 1)
inline ==1   : 0   (should be 1)
```

It happens for any operand whose width is not a multiple of 32. `$onehot` and `$countbits` have the same problem; `$display` and assignment are fine.

The VPI getter for thread-vector values reuses a scratch buffer and only fills the bits it needs, so the unused high bits of the top word keep stale data and these functions count it. Workaround: assign to a variable first.

Build: Icarus Verilog 13.0 stable and current master.

drewbabel:
Fixes #1411.

`$countones`, `$onehot`, `$onehot0`, and `$countbits` can return the wrong value when used directly in an expression (e.g. `$countones(x) == 1`) for any operand width that is not a multiple of 32. The result depends on leftover buffer contents, so it is not deterministic, and assigning to a variable first usually hides it.

The VPI getter `__vpiVThrVec4Stack::vpi_get_value_vector_` fills a reused scratch buffer over bits `0..wid-1` only, leaving stale data in the unused high bits of the most significant word. The bitvec consumers count whole 32-bit words, so they count that garbage. The other value getters already zero their buffers; this one did not.

The fix zeros the buffer before filling it. Adds `sf_countones_inline.v` covering the inline case at non-32-multiple widths (the existing test only assigns to a variable).

### DIFF
--- a/ivtest/ivltests/sf_countones_inline.v
+++ b/ivtest/ivltests/sf_countones_inline.v
@@ -1,0 +1,57 @@
+// $countones / $onehot / $countbits used inline in an expression at a width
+// that is not a multiple of 32. Regresses the stale-high-bits bug in the
+// VPI thread-vector getter. Existing sf_countones.v only assigns to a var.
+module top;
+  reg pass;
+  reg [3:0]  an;
+  reg [11:0] w12;
+
+  initial begin
+    pass = 1'b1;
+    an   = 4'b1110;   // ~an = 0001 -> exactly one set bit
+    w12  = 12'b0000_0010_0000;
+
+    // The original failing idiom: one-cold / one-hot check inline.
+    if (($countones(~an) == 1) !== 1'b1) begin
+      $display("FAILED: $countones(~an)==1 should be true (an=%b)", an);
+      pass = 1'b0;
+    end
+
+    if (($countones(~an) != 1) !== 1'b0) begin
+      $display("FAILED: $countones(~an)!=1 should be false (an=%b)", an);
+      pass = 1'b0;
+    end
+
+    // Inline in arithmetic context.
+    if (($countones(~an) + 0) !== 1) begin
+      $display("FAILED: $countones(~an)+0 should be 1, got %0d", $countones(~an) + 0);
+      pass = 1'b0;
+    end
+
+    // Width that is not a multiple of 32, inline.
+    if (($countones(w12) == 1) !== 1'b1) begin
+      $display("FAILED: $countones(w12)==1 should be true (w12=%b)", w12);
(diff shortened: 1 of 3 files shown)


caryr:
I've been looking at this and it is a bit more insidious than just impacting `$countones()` (e.g. `%u` also assumes the upper bits are cleared properly). I believe the correct solution may be to use `memset()` to zero all the allocated bytes in 'need_result_buf()' instead of just fixing `vpi_get_value()` to clear the individual words. My concern is will this noticeably impact performance? I assume `memset()` is coded to be very efficient so it should be negligible.

I think it is currently safe when using a RBUF_STR since they are NULL terminated, but RBUF_VAL is never safe if the upper bits are not masked properly. It's probably best to also clear the string value even if it doesn't impact the code. to avoid things appearing in the buffer that could be snooped.

@martinwhitaker, @larsclausen do you have an opinion?

martinwhitaker:
The allocated buffer size is rounded up to a multiple of 4Kbytes, so clearing the whole buffer is going to be costly.

I would have just cleared the last 32 bits in the buffer:
````
      unsigned hwid = (wid+31)/32;
      vp->value.vector = static_cast<s_vpi_vecval*>
                         (need_result_buf(hwid*sizeof(s_vpi_vecval), RBUF_VAL));
      assert(vp->value.vector);
      if (hwid > 0) {
            vp->value.vector[hwid-1].aval = 0
            vp->value.vector[hwid-1].bval = 0
      }
````

caryr:
Or as an alternative; `assert()` the width is greater than 0 before the `hwid` calculation and then just do the assignments to zero directly.

ETA: I still think this should go in the `need_result_buf()` code somehow, because there are likely other places we should be zeroing out the upper bits other than when filling the `vector` area, but it's just a raw bunch of bytes in `need_result_buf()` so that seems like an all or nothing case. Maybe we just need to do an audit to make sure there are no other places we are letting old values leak.

PR #1413, opened 2026-06-30
Support declaration names shadowing type identifiers

larsclausen:
SystemVerilog allows a declaration in an inner scope to use the same name as a type identifier from an outer scope. The lexer reports such names as `TYPE_IDENTIFIER` before the new declaration has been installed, which made constructs such as `int T;`, `wire T;`, and `input T` fail when `T` was a visible typedef.

The affected declaration forms have a local type/name/dimension ambiguity. For example, after `input T` or `wire T` the parser does not know whether `T` is the declared name, or whether a following identifier will make `T` the declaration type in `input T x` or `wire T x`. With dimensions, `input T [1:0]` and `wire T [1:0]` can be either a declaration named `T` with unpacked dimensions or a declaration using typedef `T` as a packed type followed by another name.

Parse these declaration forms with productions that decide the first declarator and carry the selected declaration type across the rest of the list. This covers variable declarations, net declarations, ANSI and non-ANSI module port declarations, and task/function port declarations. Other identifier uses still need separate grammar changes.


caryr:
Do we warn or have an option to warn when variable shadowing occurs? I'm 100% behind supporting what is allowed by the standard, but I'm also 100% behind providing warnings so people can avoid coding practices that can create confusion.

larsclausen:
> Do we warn or have an option to warn when variable shadowing occurs? I'm 100% behind supporting what is allowed by the standard, but I'm also 100% behind providing warnings so people can avoid coding practices that can create confusion.

There is already support for identifiers shadowing identifiers and there is no warning at the moment. This PR allows shadowing of types names by identifiers. I can add support for warnings for both as a follow up PR.

caryr:
I think having a shadow warning we could enable would be beneficial.

PR #1418, opened 2026-07-01
SV: chained method calls a().b()

muhammadjawadkhan:
Split from steveicarus/iverilog#1330 (**part 2/6**). **Depends on #1416** — merge that first.

## Summary

Adds **chained method calls** on expression results, e.g. `obj.method().other_method()`, extending the class property infrastructure from #1416.

## User-visible behavior

- Parse and elaborate **chained calls** `a().b()` on function/method return values.
- Works with the class queue/darray property paths introduced in #1416.

## Implementation notes (high level)

- **Parser / pform:** `PECallFunction` chaining rules in `parse.y`, `pform.cc`.
- **Elaboration:** `elab_expr.cc` — `elaborate_class_method_net_` for chained class method calls.

## Tests / regressions

- `sv_call_chain_method1` — chained call regression.

## How to verify

```text
make -j$(nproc)

iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_call_chain_method1.v && vvp sim.vvp
```

Requires #1416 merged or this branch checked out (includes part 1).

## Commits

**1** new upstreamable commit on top of #1416:

- `SV: chained method calls a().b()`

---

## Reviewer checklist

- [ ] Chained calls parse and elaborate without errors
- [ ] `sv_call_chain_method1` passes simulation
- [ ] No regressions in class property queue tests from #1416

PR #1416, opened 2026-07-01
SV: class queue/darray property foundation

muhammadjawadkhan:
Split from steveicarus/iverilog#1330 (**part 1/6**). Replaces the monolithic #1330 with a stacked series; merge **#1416 → #1418 → #1419 → #1420 → #1421 → #1422** in order. (#1417 is independent.)

## Summary

This PR lays the **foundation** for SystemVerilog **class properties** whose type is a **queue** or **dynamic array**. It wires parsing/elaboration through **tgt-vvp** and **VVP**, extends **VPI** / `%p` display, and fixes **opcode_table** ordering so class queue property opcodes remain valid for VVP's `bsearch` lookup.

Later parts of the series build on this branch for chained calls, locator methods, reductions, and ordering.

## User-visible behavior

- **Class fields** of type queue or dynamic array: assignment, indexing, and built-in queue methods (`push_back`, `pop_front`, etc.) on class properties.
- **`%p`** and VPI can display class queue/darray property values where implemented.
- No new locator / reduction / ordering methods in this PR — those arrive in #1418–#1422.

## Implementation notes (high level)

- **Elaboration:** `NetEProperty` paths for queue/darray class members; `elaborate_queue_method_` accepts property expressions.
- **tgt-vvp / VVP:** object/property opcodes for queue and darray class fields; `vthread` dispatch and `vpi_darray` helpers.
- **Opcode table:** keep entries sorted for `bsearch` (class queue property operations).
- **VPI / targets:** `%p` and draw paths for class queue/darray properties.

## Tests / regressions

- `sv_class_queue_prop_methods` — class queue property push/pop methods.

## How to verify

```text
make -j$(nproc)

iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_class_queue_prop_methods.v && vvp sim.vvp
```

Run the vvp regression suite you normally use (e.g. `ivtest/vvp_reg.py` with `regress-vvp.list`).

## Commits

**1** upstreamable commit on `upstream/master`:

- `SV: class queue/darray property foundation`

---

## Reviewer checklist

- [ ] Class queue/darray properties assign and simulate as in `sv_class_queue_prop_methods`
- [ ] Queue methods on class properties elaborate and run correctly
- [ ] `%p` / VPI display behaves for covered cases
- [ ] No `opcode_table` ordering regression (VVP binary search)
- [ ] ivtest / vvp regressions pass with built `lib/ivl`

PR #1330, opened 2026-04-29
Add multiple Sytemverilog supports

muhammadjawadkhan:
## Summary

This series extends Icarus’s SystemVerilog support for **class properties** that are **queues** and **dynamic arrays**: parsing/elaboration, **tgt-vvp** and **VVP** runtime, **VPI**, chained calls, and **ivtest** regressions. It adds **locator-style methods** (`find*`, `unique*`, `min` / `max` with optional `with` predicates), **numeric reductions** (`sum`, `product`, including expressions), and **ordering** (`reverse`, `sort`, `rsort`, `shuffle`) for queues and dynamic arrays, including **class property** paths.

## User-visible behavior

- **Class fields** of type queue / dynamic array: assignment, method-style usage where implemented, and integration with existing SV features (`%p`, property chains).
- **Method chains / chained calls** on expressions (e.g. `a().b()`), with a regression.
- **Locator methods** on queues and dynamic arrays: `find`, `find_index`, `find_first`, `find_last`, and related index variants; `unique`, `unique_index`; `min`, `max`, with **`with`** predicates where applicable; extension to **class queue** and **class dynamic-array** properties.
- **Reductions:** integral **`sum()`** and **`product()`** on queues and dynamic arrays; operands may be **expressions** (not only simple identifiers).
- **Ordering:** **`reverse()`**, **`sort()`**, **`rsort()`**, **`shuffle()`** on queues and dynamic arrays (including class property targets).
- **VVP:** array-pattern handling for **object** values in **dynamic-array** and **class queue** property assignments where added in this series.

## Implementation notes (high level)

- **Elaboration / netlist:** new or extended paths for the above methods; synthesis checks updated where methods are non-synthesizable (e.g. ordering, like existing `delete` / `reverse` style warnings).
- **tgt-vvp / VVP:** opcodes and `vthread` dispatch aligned with existing **object** / **property** patterns (queue vs dynamic array, including **prop** opcodes for class fields); **`vvp_darray`** helpers for ordering and reductions; object ordering uses a stable **sort key** for class handles where needed.
- **Opcode table:** fix so **`opcode_table` remains correctly ordered for `bsearch`** (class queue property operations).
- **VPI / targets:** groundwork for class queue/darray properties and related codegen paths (as in the branch).

## Tests / regressions

[...]

PR #1417, opened 2026-07-01
Windows: route vpip_format_pretty through vpip_routines_s

muhammadjawadkhan:
Split from steveicarus/iverilog#1330 (**independent** of the #1416–#1422 feature stack). Can merge in any order relative to those PRs.

## Summary

Fixes **MinGW / Windows** link failures when building VPI modules: route **`vpip_format_pretty`** through **`vpip_routines_s`** (version bump) and export **`ivl_type_queue_max`** in **`ivl.def`**.

## User-visible behavior

No functional change on Linux. Windows CI / MSYS2 builds link `system.vpi` and related targets without undefined references to `vpip_format_pretty`.

## Implementation notes (high level)

- **`vpi_user.h`:** bump `vpip_routines_version` 1→2; add `format_pretty` to `vpip_routines_s`.
- **`vpi/libvpi.c`:** MinGW thunk `vpip_format_pretty()` → `vpip_routines->format_pretty`.
- **`vvp/vpi_priv.cc`:** jump table entry for `format_pretty`.
- **`ivl.def`:** export `ivl_type_queue_max` for MinGW.

## Tests / regressions

No new ivtest. Verified by Windows CI link of `system.vpi` (was failing with undefined reference to `vpip_format_pretty`).

## How to verify

```text
make -j$(nproc)
# On MSYS2 / MinGW: build should complete without vpip_format_pretty linker errors.
```

## Commits

**1** upstreamable commit on `upstream/master`:

- `Windows: route vpip_format_pretty through vpip_routines_s`

---

## Reviewer checklist

- [ ] `vpip_routines_version` bump is consistent across VPI consumers
- [ ] MinGW thunk and `vpi_priv.cc` jump table match
- [ ] `ivl.def` export list complete for Windows
- [ ] Linux build unaffected

PR #1419, opened 2026-07-01
SV: queue/darray locator methods (find*, unique*, min/max)

muhammadjawadkhan:
Split from steveicarus/iverilog#1330 (**part 3/6**). **Depends on #1416, #1418** — merge those first.

## Summary

Adds **locator-style methods** for plain **queues** and **dynamic arrays**: `find*`, `unique*`, `min` / `max`, including **`with`** predicates where applicable. Includes VVP runtime support (`eval_object.c`, new opcodes, `vthread` dispatch).

## User-visible behavior

- **Locator methods** on queues and dynamic arrays:
  - `find`, `find_index`, `find_first`, `find_last`, and related index variants
  - `unique`, `unique_index`
  - `min`, `max`, with optional **`with`** predicates
- Class **property** locator paths are **not** in this PR — see #1420.

## Implementation notes (high level)

- **Elaboration:** `elab_expr.cc` — `NetESFunc` / iterator-scope handling for `with` predicates.
- **tgt-vvp / VVP:** `eval_object.c`, opcode additions, `vthread` reduction/locator dispatch.
- **Parser:** `with`-clause on array method calls where needed.

## Tests / regressions

- `sv_queue_find`, `sv_queue_find_locators_ext`, `sv_queue_find_with`
- `sv_queue_unique`
- `sv_darray_find_locators`
- `README_sv_queue_locators.txt` (introduced/extended in this part)

## How to verify

```text
make -j$(nproc)

iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_queue_find.v && vvp sim.vvp
iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_darray_find_locators.v && vvp sim.vvp
```

## Commits

**1** new upstreamable commit on top of #1418:

- `SV: queue/darray locator methods (find*, unique*, min/max)`

---

## Reviewer checklist

- [ ] Locator methods return expected indices/values for queue and darray tests
- [ ] `with` predicates filter elements correctly
- [ ] `min` / `max` match expected ordering for supported element types
- [ ] ivtest / vvp regressions pass


muhammadjawadkhan:
Rebased and squashed: each branch now has clean, upstreamable commit(s) on top of `upstream/master`.

- **#1416 / split/01:** 1 commit
- **#1418 / split/02:** 1 new commit (2 total on branch)
- **#1419 / split/03:** 1 new commit (3 total)
- **#1420 / split/04:** 1 new commit (4 total)
- **#1421 / split/05:** 1 new commit (5 total)
- **#1422 / split/06:** 1 new commit (6 total)
- **#1417 / split/07:** 1 commit (independent)

Each PR adds exactly one presentable commit; merge in order #1416 → #1422.

caryr:
See my previous comment regarding `if` formatting in the other pull request.

muhammadjawadkhan:
Applied, single-line early returns where appropriate, { } on multi-line if bodies, and || at end of line with aligned continuations. Updated in elab_expr.cc, eval_object.c, netmisc.cc, and vthread.cc.

caryr:
Thanks, I'll try to review this later today.

caryr:
Just a few things.

How is all this going to scale? You have added a bunch of code, but it all appears to be queue specific. How will this be altered to add supported for the other array and value types? I'm concerned this is heading in the direction of just adding code, but is lacking a coherent underlying design that factors in the requirements of all that should be supported in the future. Can you describe your plan. Does it make sense for any of this to be a system function written in VPI instead of adding new opcodes to support everything? I'm not making any specific suggestions just trying to see if the appropriate questions and design alternatives have been explored.

Why was the mode functionality not implemented as an enumeration? That gives the numeric functionality along with named constants that can be used to make the code clearer instead of using plain numeric constants.

Why was the comment for `draw_darray_pop()` removed in tgt-vvp/eval_vec4.c? That routine was not changed.

There are a few places where `&&` in an `if` is not at the end of the line continued line.

muhammadjawadkhan:
This PR deliberately follows the same path as #1416 (push/pop/size): elaborate queue/darray locator calls to internal $ivl_queue_method$* system functions, lower them in tgt-vvp, and execute via VVP opcodes in vthread. It is not meant to be queue-only long term, but the first slice is intentionally narrow: integral packed elements (bool/logic) on queues and unpacked dynamic arrays, plus class-property paths via the prop/v opcode variants from #1416.

For runtime sharing, the hot loops are template helpers (queue_run_unique_src, queue_run_find_matches_src, queue_find_first_last_to_queue_src) parameterized over vvp_queue_vec4* / vvp_darray*, so queue vs darray does not duplicate logic. Property vs plain signal is handled by separate opcodes (again matching existing queue-method conventions), not separate algorithms.

What is not covered yet (and where growth goes): real, string, and class-element queues/darrays. The elaboration side already gates unsupported element types with explicit errors. Extending support would mean:

typed lowering in tgt-vvp (as eval_vec4.c / eval_string.c / eval_real.c already do for pop_back/pop_front), and
typed runtime helpers or opcode families, reusing the same template iteration pattern where comparison semantics are shared.
I did consider VPI system functions. The existing queue method infrastructure in Icarus uses compile-time lowering + opcodes because the simulation path is hot and needs tight integration with the VVP stack (especially for with, which must inject item/index into a scope and branch on a compiled predicate). A VPI implementation would still need similar iteration/compare logic at runtime, but with more overhead and without an established pattern in this codebase for queue iterators. So opcodes match how #1416 and upstream queue methods already work; VPI might be reasonable for rarely-used or debugging-oriented helpers, but did not look like the right default here.

Later PRs in the #1330 split (sum/product, sort/shuffle, class-property locators) build on this same $ivl_queue_method$* + opcode model rather than adding a parallel mechanism.

Mode as enumeration

Agreed, the numeric mode in eval_queue_method_find_with() was a quick way to unify six *_with variants in one code generator. I'll refactor it to a named enum so the find/find_index/first/last × forward/reverse cases are explicit.

draw_darray_pop() comment

That was an accidental deletion when adding the nearby find*_with dispatch in eval_vec4.c; draw_darray_pop() itself was not changed. I'll restore the original comment.

I'll fix the remaining multi-line if conditions so && stays at the end of the continued line, per your #1416/#1418 convention.

muhammadjawadkhan:
Addressed the concrete items, restored the draw_darray_pop() comment, refactored find*_with mode to a named enum with small helpers, and fixed the remaining && continuation formatting in elab_expr.cc. Design/scaling rationale is in my earlier comment , i am happy to discuss further if you want changes before merge.

caryr:
In general please think about how this can all be refactored to avoid an explosion of opcodes. to support all types.

A couple more things to address before merging:

1. In `tgt-vvp/eval_object.c` and `elab_expr.cc` it looks like there is a huge amount of code duplication. It looks like you should be able to refactor much of this code into a couple routines that are parameterized appropriately.
2. There are a number of places where you are not indenting routine arguments so they are at the same level as the next statements and this makes it complicated to visually see the arguments. This should be done consistently following what the code currently does.

muhammadjawadkhan:
Thanks for pointing out these changes.

1. Deduplication
Collapsed the repeated unique*/find* elaboration in elab_expr.cc into a single parameterized helper (elab_array_locator_method) used for class-property queues, plain queues, and dynamic arrays. In tgt-vvp/eval_object.c, property vs signal lowering for find/unique is shared via emit_queue_locator_opcode, with table-driven opcode selection, and the find*_with property/var paths share init/loop/match helpers instead of duplicated bodies.

2. Argument indentation
Multi-line call/constructor arguments are aligned consistently with surrounding Icarus style.

Opcode growth
This slice still uses vec4-specific opcodes (matching existing queue methods), but the structure is set up so additional element types should extend through typed helpers / parameterized opcodes rather than cloning whole method blocks. Happy to discuss further consolidation (e.g. fewer opcodes with an element-type operand) if you want that before merge.

PR #1422, opened 2026-07-01
SV: reverse(), sort(), rsort(), shuffle() for queues/darrays

muhammadjawadkhan:
Split from steveicarus/iverilog#1330 (**part 6/6**). **Depends on #1416–#1421** — merge those first.

## Summary

Adds **ordering methods** for queues and dynamic arrays: **`reverse()`**, **`sort()`**, **`rsort()`**, **`shuffle()`**, including class property targets. Updates ivtest gold files and regression lists per earlier review feedback.

## User-visible behavior

- **`reverse()`** — reverse queue/darray elements in place.
- **`sort()`** / **`rsort()`** — sort ascending/descending (stable sort key for class handles where needed).
- **`shuffle()`** — random permutation.
- Synthesis emits warnings for non-synthesizable ordering methods (consistent with existing `delete` / `reverse` style checks).

## Implementation notes (high level)

- **Elaboration / netlist:** `elaborate.cc`, `netlist.cc` — ordering/delete method warnings.
- **tgt-vvp / VVP:** `vvp_process.c`, `vvp_darray.cc` helpers, opcodes, `vthread` ordering dispatch.

## Tests / regressions

**New tests:**
- `sv_queue_reverse`, `sv_queue_sort`, `sv_darray_reverse`, `sv_darray_sort`
- Class property locator regressions extended for ordering

**Review-driven ivtest updates:**
- `always_comb_warn`, `always_ff_warn`, `always_latch_warn` gold — ordering/delete method message
- `br1005` gold / `regress-vlog95.list` — class queue compiles (was CE)
- `br_gh710b` — restore LXT args/gold per reviewer

## How to verify

```text
make -j$(nproc)

iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_queue_sort.v && vvp sim.vvp
iverilog -g2005-sv -o sim.vvp ivtest/ivltests/sv_darray_reverse.v && vvp sim.vvp
```

Full regression (as run locally before split):

```text
cd ivtest
perl vvp_reg.pl
perl vpi_reg.pl --with-pli1
python3 vvp_reg.py
python3 vvp_reg.py --vlog95
perl vlog95_reg.pl
```

## Commits

**1** new upstreamable commit on top of #1421:

- `SV: reverse(), sort(), rsort(), shuffle() for queues/darrays`

---

## Reviewer checklist

- [ ] `reverse` / `sort` / `rsort` / `shuffle` behave as in regressions
[...]

PR #1446, opened 2026-07-21
SV: finish array methods for sv-tests section 7.12

muhammadjawadkhan:
## Summary
- Follow-up to #1422 for caryr’s [sv-tests section 7.12](https://chipsalliance.github.io/sv-tests-results/) request (prior commit landed after that PR merged).
- No-parens darray locators and string-element `find* with` (assign locator results to queues).
- `and()` / `or()` / `xor()` array reductions (LRM 7.12.3).
- Iterator `item.index` in `with` predicates (LRM 7.12.4).

## Test plan
- [x] Local chipsalliance sv-tests for 7.12 / 7.12.1 / 7.12.2 / 7.12.3 / 7.12.4 all pass
- [x] ivtest: `sv_darray_and_or_xor`, `sv_darray_item_index`, plus related locator/ordering/sum tests


caryr:
Make sure to read my latest comment regarding the three failing tests in #1422. I'll look at this pull request tomorrow.

muhammadjawadkhan:
Added a small init-hardening commit for caryr’s report on #1422 (local segfaults on the reverse/sort/class-prop tests that CI does not see):

- Initialize all thread flags + word registers in `vthread_new`
- Value-initialize VVP codespace chunks

I could not reproduce under Valgrind/ASan locally; this is defensive cleanup around the “uninitialized value” suspicion.

caryr (commented):
This is not needed and shouldn't be done. We want uninitialized values to be tagged by valgrind. I still don't know what the issue is, but a make clean and rebuild solves the issue which explains why the CI works and my local builds on the same machine style fail. My guess is a make to the previous head and then adding the new code will allow you to reproduce the dependency issue. Both gcc and clang are generating code that fails so it's likely not a compiler bug.

muhammadjawadkhan:
Reverted the init-hardening commit (`621f589e0`) — agreed that forcing zero-init hides Valgrind findings.

I was able to reproduce the three segfaults with your incremental-build recipe:

1. `make clean` + full build at `f0b6d3add` (pre-ordering / pre-#1422)
2. Check out current tip **without** clean
3. Incremental `make` + install
4. `sv_darray_reverse`, `sv_darray_sort`, `sv_class_darray_prop_locators` all **SIGSEGV**

GDB on the broken `vvp`:

```
#0  0x0000000000000000 in ?? ()
#1  of_REVERSE_OBJ at vthread.cc:2629
#2  vthread_run at vthread.cc:932
```

So it is a call through a **NULL** (broken vtable / mismatched object layout), not a random uninit flag. After `make clean` + rebuild on the same tree, all three tests **PASSED**.

That matches CI vs your local machine: clean builds are fine; stale objects after adding virtuals on `vvp_darray` (reverse/sort/shuffle) break incremental links. I will not re-land the init commit.

caryr:
Given this is indicated as a failure in calling a queue method and the queue base class does not explicitly override the reverse, sort or shuffle methods it inherits from the array base class and these are then overridden in the sub-classes of the queue base class maybe that is a clue. Of course I cannot answer why a clean rebuild solves the problem.

caryr:
Did you intend to close this? There are still items to be merged. I've been out of town for a bit so am just getting back to looking at this now.

PR #1447, opened 2026-07-21
SV: class-handle $cast and static $typename

muhammadjawadkhan:
## Summary
- Implement class-handle `$cast(dest, src)` (function + task) with runtime is-a checks along the class inheritance chain (`%cast/cobj`).
- Fold `$typename` to a documented static type string (classes → `class <name>`).
- Example `examples/cast_typename` prints **PASSED**; docs in `docs/cast-typename.md`; STATUS/ROADMAP updated.

## Test plan
- [x] `cd examples/cast_typename && make run` → `typename=class derived` + `PASSED`
- [ ] Rebuild from clean `./configure && make && make install` and re-run the example
- [ ] Spot-check failed cast (base-only object into derived) returns 0 and leaves dest unchanged


muhammadjawadkhan:
Opened on wrong remote; this work targets muhammadjawadkhan/iverilog-uvm only.

Issue #1149, opened 2024-07-23
Unpacked localparam array support


lefp:
As far as I can tell, IEEE 1800-2012 supports unpacked arrays as localparams, and the syntax used in the following example.

However, the following fails to compile:

```verilog
module Asdf ();

    localparam int i[3] = { 1, 2, 3 };

endmodule
```
```
> iverilog -g2012 Asdf.sv
Asdf.sv:3: error: localparam must have a value.
Asdf.sv:3: syntax error
Asdf.sv:3: error: Invalid module item.
```

## Related issues and work

There is an issue #562 about array assignment in general, which was partially addressed by #953. That seems to have solved assignment to variables, but it doesn't work for localparams.

Unpacked and multidimensional localparam support was also mentioned in discussion in #570, but since that issue was focused on function-based assignment, I have opened this one as a separate issue.

caryr:
As written this still fails after the recent change to add support for unpacked array parameters, etc.

Though if you change the unpacked array concatenation (still not supported) to an array assignment pattern (is supported) it works as expected.

Issue #1180, opened 2024-11-07
Cryptic error messages when trying to use packed and unpacked array parameters


cousteaulecommandant:
On this issue I'm reporting:
* A bug(?) I found when trying to declare a parameter as a *packed* array (note that issue #846 refers to *unpacked* arrays).
* A suggestion/request to improve the error message displayed on unsupported features.
________

I was trying to simulate this code (well, not this code exactly but a similar one), which seems to use an unsupported feature (unpacked array parameters):
```sv
module TEST_param_arr;

parameter int num = 6;
parameter int vars [0:num-1] = '{4, 8, 15, 16, 23, 42};

initial begin
    for (int i=0; i<num; i++)
        $display("vars[%0d] = %2d", i, vars[i]);
    $finish;
end

endmodule
```
and got this rather cryptic error message:
```
TEST_param_arr.sv:4: error: parameter declared outside parameter port list must have a default value.
TEST_param_arr.sv:4: syntax error
TEST_param_arr.sv:4: error: invalid module item.
```
I managed to figure out that this was due to issue #846 *(Parameters having an **unpacked** array type is a SystemVerilog feature that isn't supported yet)*, but iverilog could at least try to figure out what I was trying to do and tell me that that's not supported; the error message as it is is quite confusing.

Now, since a comment in that issue mentions that I cannot use *unpacked* arrays, I thought OK, I'll use a *packed* array instead:
```sv
module TEST_param_arr2;

parameter int num = 6;
parameter [0:num-1] [31:0] vars = {32'd4, 32'd8, 32'd15, 32'd16, 32'd23, 32'd42};

initial begin
    for (int i=0; i<num; i++)
        $display("vars[%0d] = %2d", i, vars[i]);
    $finish;
end

endmodule
```
and got this much more colorful error message:
```
TEST_param_arr2.sv:8: assert: ../iverilog/netmisc.cc:1821: failed assertion packed_dims.size() == 1
```
I honestly don't know how to interpret that error message, other than it seeming to be related to the use of packed dimensions.
In this case, the issue seems to be in the *use* of the variable though; for some reason using a single index on a multidimensional packed parameter.  I tried replacing `vars[i]` with `vars[i][31:0]` with no luck.

[...]

Issue #570, opened 2021-12-20
Initialize array via function within package


logicmonkey:
```
package mypkg;
  localparam ELEM = 4;
  typedef int myints_t [ELEM-1:0]; // create a type that is an array of ints

  localparam myints_t myarr = init_arr(ELEM); // initialise constant as an array of values set by a function

  function automatic myints_t init_arr(int N); // an easy function
  begin
    for (int i=0; i<N; i++)
      init_arr[i] = i;
  end
  endfunction
endpackage

module dummy;
import mypkg::*;
  initial begin
    for (int i=0; i<mypkg::ELEM; i++)
      $display(mypkg::myarr[i]);
    $display(mypkg::ELEM);
    $finish;
  end
endmodule
```

Expected output (as per commercial simulator):
```
xcelium> run
          0
          1
          2
          3
          4
Simulation complete via $finish(1) at time 0 FS + 0
```

Actual output (some rev 12 hash build run with iverilog -g2012)
```
0
1
0
1
          4

```

martinwhitaker:
Icarus Verilog does not currently support unpacked arrays as (local) parameters or function return values, or indeed, most other places where they can't be used in traditional Verilog. But it should warn you about this and fail to compile your code, rather than producing incorrect results.

logicmonkey:
I tried enforcing PACKED types - same problem (ATTEMPT1)

```
package mypkg;
  localparam ELEM = 4;

  typedef logic  [$clog2(ELEM)-1:0] elem_t;     // pack this
  typedef elem_t [ELEM-1:0]         mypints_t;  // and this

  localparam mypints_t ATTEMPT1 = init_arr(ELEM);  // initialise constant as an array of values set by a function

  function automatic mypints_t init_arr(int N); // an easy function
  begin
    for (int i=0; i<N; i++)
      init_arr[i] = i;
  end
  endfunction

  //typedef int myints_t [ELEM-1:0];                // unpacked array of ints
  //localparam myints_t ATTEMPT2 = '{3, 2, 1, 0};

  //localparam logic [ELEM-1:0][$clog2(ELEM)-1:0] ATTEMPT3 = 8'b11_10_01_00;

endpackage

module dummy;
import mypkg::*;
  initial begin
    for (int i=0; i<mypkg::ELEM; i++)
      $display(mypkg::ATTEMPT1[i]);

    //for (int i=0; i<mypkg::ELEM; i++)
    //  $display(mypkg::ATTEMPT3[i]);

[...]

PR #501, opened 2021-04-20
Preliminary unpacked support for localparam

phhorrein:
As we definitely need this in our projects, I've started working on addition of unpacked array support initialization. 

This is a first try at doing this for localparam only, to see how it goes. I've been focusing on "getting it to work", and thus a lot of things need to be added. At least, the vpi_draw and t-dll file must be updated to take the new uarray into accounts.

However, I would really like to contribute this to the project if you're interested, and thus I want to discuss how I did this before going further. So here is a draft pull request with a skeleton of how I have done it.

As this is my first meaningful modification of Icarus source code, I would really appreciate any feedback you can give, on the design choices or on the coding style, or on whole parts of the code I missed. I will do my best to answer them and to modify the code. 

I have small tests I added manually to ivtest, I can push them on a separate PR if you want.

PR #1450, opened 2026-07-26
Support unpacked arrays as parameters, constant-function returns, and subroutine ports

### LINKED ISSUE #1449: Assertion failure (SIGABRT) in eval_tree.cc for constant function returning an unpacked array
## Summary

`iverilog` aborts with an internal assertion, rather than emitting a diagnostic, when a constant
function returns an unpacked array and the result initialises a `localparam`.

```
foo.sv:16: assert: eval_tree.cc:217: failed assertion lval.len() == wid
Aborted
```

Exit status is 134 (128 + SIGABRT). Reproduces on current master, `s20260301-286-ga4989d023`.

## Self-contained example

```systemverilog
module top;
  typedef logic [9:0] exp_t [0:7];

  function automatic exp_t build_exp();
    exp_t t;
    logic [9:0] a = 10'd1;
    for (int i = 0; i < 8; i++) begin
      t[i] = a;
      a = a + 10'd1;
    end
    return t;
  endfunction

  localparam exp_t EXP = build_exp();

  initial begin
    $display("%0d", EXP[3]);
    $finish;
  end
endmodule
```

```sh
iverilog -g2012 -o a.out foo.sv
```

## Expected

Either the parameter is evaluated (Riviera-PRO, Xcelium and Vivado all accept this shape), or a
clean `sorry:` / `error:` diagnostic. Not a crash.

## Actual

```
foo.sv:16: assert: eval_tree.cc:217: failed assertion lval.len() == wid
Aborted
```

## Context

This started as a real-world case: Reed-Solomon decoders over GF(2^10) whose field package builds
log / antilog / inverse ROMs with elaboration-time constant functions returning unpacked arrays
(`typedef logic [9:0] rom_t [0:1023];`). The same package shape produces the cleaner
`the type of the variable 'r' doesn't match the context type` error on most tables, but this one
aborts instead. Both come from the same root cause: a vector-width assumption applied to an
unpacked-array lvalue in the constant-function evaluator.

Related, and part of why I ran into it:

- #1149 and #570 are the open feature requests for this construct.
- #1180 was closed by adding `sorry` messages for array parameters, which improved the diagnostics
  but does not cover this path, which still asserts.
- #501 is a stalled PR for the feature; a 2021 comment there hit the same `lval.len() == wid`
  assertion, but against a patched build. On stock master it seems unreported.

## Fix

I have a patch and will open a PR shortly. It removes the crash by implementing the construct, which
also closes #1149 and #570. If you would rather not take the feature, the minimal fix for this issue
alone is to turn that assertion into a diagnostic.

## Version

```
Icarus Verilog version 14.0 (devel) (s20260301-286-ga4989d023)
```
Ubuntu 24.04 on WSL2, g++ 15.2.0.

marcobrambilla99:
Makes unpacked arrays usable as values: as parameters, as the return of a constant function, in a whole-array assignment, and as subroutine ports.

Fixes #1449 (SIGABRT), and implements #1149 and #570.

## What it does

| Form | Example |
|---|---|
| 1-D unpacked parameter | `localparam logic [3:0] MEM [0:63] = '{...}` |
| N-D packed parameter | `localparam logic [0:63][3:0] MEM = '{...}` |
| N-D unpacked parameter | `localparam logic [3:0] TBL [0:3][0:15] = '{'{...},...}` |
| constant function returning an unpacked array | `function automatic rom_t f(); rom_t r; ...; return r; endfunction` |
| whole-array assignment at run time | `b = a;` for `logic [7:0] a [0:3], b [0:3]` |
| unpacked array as a subroutine port | `function automatic logic [7:0] sum4(input logic [7:0] a [0:3]);` |

An N-D unpacked value is flattened row-major into a single packed constant by recursing through the
nested `'{...}` pattern. An index select that consumes every unpacked dimension yields the packed
element, for constant and variable indices and for ascending or descending declared ranges.

## Relationship to existing work

- **#1449** is a crash: on master, a constant function returning an unpacked array can abort with
  `eval_tree.cc:217: failed assertion lval.len() == wid` rather than diagnosing. That is the same
  root cause as the `doesn't match the context type` error, a vector-width assumption applied to an
  unpacked-array lvalue, and the same commit fixes both.
- **#1180** was closed by adding `sorry:` messages for array parameters. This PR necessarily removes
  those, since the construct now works, and inverts the two regression tests that asserted them
  (`br_gh1180a`, `br_gh1180b` move from gold-stderr to normal). To be clear, that is not undoing
  deliberate work: #1180 asked for better error messages, and the feature requests it pointed at,
  #1149 and #570, are still open.
- **#501** is the earlier attempt at this, still open as a draft from 2021. To be accurate about why
  it stopped: the `IVL_VT_UARRAY` design question was raised and resolved within two days, and the
  author removed the type himself. What it never got through was a long tail of crashes that
  @jlwehle reported against the branch through mid-2021, each a different assertion. In 2023 he
  offered to finish the work and the thread went quiet.

  This PR also adds no new type, keeping unpacked arrays as special cases of the base types. More
  usefully, the shapes that stalled #501 are covered here and pass with correct values, not merely
  compiling: an element of one array parameter feeding another's initialiser (nested two levels), an
  element inside `$clog2`, `integer` rather than logic elements, and a run-time variable index in
  both a continuous assignment and a procedural block. `param_array_param_ref` pins all of them,
  including the `lpa_tc` case attached to that thread.

## Conflict note

`elab_expr.cc` needed a manual merge against the recently merged queue-method work. Both changes are
kept: the queue-method dispatch returns early for queue types, and the new whole-unpacked-array type
resolution runs after it. The `netdarray_t` compatibility check now tests
`array_type->type_compatible(net_ctype) && sr.path_tail.empty()`, combining both conditions.

## Testing

Against master `a4989d023`, x86-64 Linux, g++ 15.2.0:

```
master:  vvp_reg.py  519 ran, 3 failed
this PR: vvp_reg.py  528 ran, 3 failed
```

The 9 extra are new tests added here, all passing. The 3 failures are `br_gh710b`, `br_gh710c` and
`br_gh710d`, which fail identically on unmodified master, so there are no regressions.

New tests: `br_gh1180c`, `param_array_packed`, `param_array_unpacked`, `func_return_uarray`,
`func_return_uarray_2d`, `func_return_uarray_rt_fail`, `uarray_subroutine_port` (function and task
arguments, ascending and descending ranges, two dimensions, real elements, x propagation, run-time
array copy, use from `always_comb`) and `uarray_ca_fail` (pins the continuous-assignment
diagnostic) and `param_array_param_ref` (the #501 shapes).

Also verified on the case that prompted this: Reed-Solomon decoders over GF(2^10) whose field
package builds log / antilog / inverse ROMs this way. Before, they either failed to elaborate or
aborted; after, they elaborate and produce the same golden results as commercial simulators.

## Subroutine ports

`sorry: Subroutine ports with unpacked dimensions are not yet supported` is also gone. This is legal
SystemVerilog (A.2.7 `tf_port_item` permits unpacked dimensions on a formal; 7.6 defines the
assignment) and other tools accept and synthesise it.

The run time already had the hard part: automatic, per-invocation unpacked arrays work today,
including under recursion, so no new storage machinery and no new type were needed. What was missing
was marshalling:

- `tgt-vvp`: a new `draw_array_copy()` copies word by word using the existing
  `%ix/load` + `%load/vec4a` + `%store/vec4a` idiom that `draw_array_pattern` already uses
  (`%load/ar` / `%store/reala` for real elements). There is no bulk copy opcode; one could be added
  later purely as an optimisation.
- It is used for a whole-array r-value in both the vector and real assignment paths, and for array
  arguments in `draw_ufunc.c`. An array actual is a signal reference rather than a stack value, so
  it is copied in the send pass reading the source signal directly, preserving the existing
  two-pass evaluate-then-send ordering.
- `elab_expr.cc` and `elaborate.cc`: an unpacked-array formal takes the array type as its context,
  not the element type. `net_type()` returns the element type, which is what produced
  `Array a needs an array index here` at every call site. Same fix for functions and tasks.
- `net_nex_input.cc`: when removing a function's own inputs from an `always_comb` sensitivity set,
  iterate every pin. An array port has one per word, and the code asserted a single pin.

Shape checking needed no new code: the existing context-type comparison already rejects a size or
element-type mismatch with the usual `doesn't match the context type` error.

## Still not covered

Two cases remain, both now reported cleanly rather than asserting or mis-generating:

- Passing an unpacked array to a function **from a continuous assignment**. A `.ufunc` node is
  driven by its input nexuses and an unpacked array is memory rather than a net, so it would also
  need the node to be sensitive to a change of any word, which the structural netlist cannot express
  today. The message points at `always_comb`, which does work.
- **Returning** an unpacked array from a function at run time, as opposed to in a constant context.

## Note on the commit sequence

The commits are kept separate and each leaves the tree in a valid state. One consequence worth
flagging so it does not look like churn: "Support constant functions that return unpacked arrays"
adds a clean `sorry` for a run-time whole-array assignment, and the following commit implements that
case and removes the message again. Happy to squash if you would prefer.

---

Disclosure: generative AI assisted with the rebase onto current master, the conflict resolution and
this description. I have reviewed, built and tested the result myself.

### DIFF
--- a/ca.log
+++ b/ca.log
@@ -1,0 +1,2 @@
+ca.sv: No such file or directory
+Preprocessor failed with 1 error(s).
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -5048,9 +5048,17 @@ unsigned PEIdent::test_width(Design*des, NetScope*scope, width_mode_t&mode)
       }
 
       // The width of a parameter is the width of the parameter value
-      // (as evaluated earlier).
-      if (sr.par_val != 0)
-	    return test_width_parameter_(sr.par_val, mode);
+      // (as evaluated earlier). The exception is an indexed unpacked-array
+      // parameter: its stored value is the whole (flattened) array, but an
+      // element select has the width of a single element, which is computed
+      // from the resolved element type below.
+      if (sr.par_val != 0) {
+	    bool unpacked_elem_sel = use_sel != index_component_t::SEL_NONE
+		  && dynamic_cast<const netsarray_t*>(sr.type)
+		  && !dynamic_cast<const netparray_t*>(sr.type);
+	    if (!unpacked_elem_sel)
+		  return test_width_parameter_(sr.par_val, mode);
+      }
 
       // If the identifier has a type take the information from the type
       if (type) {
@@ -6011,29 +6019,32 @@ static NetExpr* scale_expr_by_const(NetExpr*expr, unsigned long val)
  * constant index we fold it directly, otherwise we build an indexed part
  * select over a NetEConstParam. Only the outermost dimension is indexed here
  * (the common ROM/LUT case, `name[idx]`); the selected element keeps the
- * remaining packed dimensions as its width.
+ * remaining dimensions as its width.
+ *
+ * The flat bit layout differs between packed and unpacked array parameters:
+ *   - packed (packed_layout true): the value comes from a concatenation, so
+ *     the element at index sel_v sits at bit (sel_v - lsb)*elem_w for an
+ *     ascending-LSB range, mirroring an ordinary packed part select.
+ *   - unpacked (packed_layout false): the value was flattened lowest-array-
+ *     index first, so the element at index sel_v sits at bit
+ *     (sel_v - min_index)*elem_w regardless of the declared range direction.
  */
 NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 					      const NetScope*found_in,
-					      const netvector_t*par_vec,
+					      long msb0, long lsb0,
+					      unsigned long elem_w,
+					      bool packed_layout,
 					      NetExpr*sel,
 					      perm_string name) const
 {
-      const netranges_t&pdims = par_vec->packed_dims();
-      long msb0 = pdims[0].get_msb();
-      long lsb0 = pdims[0].get_lsb();
-      unsigned long outer_w = (msb0 >= lsb0) ? (unsigned long)(msb0 - lsb0 + 1)
-					     : (unsigned long)(lsb0 - msb0 + 1);
-      unsigned long full_w = par_vec->packed_width();
-
-	// The element width is the flat width divided by the number of
-	// elements in the outermost dimension.
-      unsigned long elem_w = full_w / outer_w;
+      long min_idx = (msb0 <= lsb0) ? msb0 : lsb0;
+      long max_idx = (msb0 <= lsb0) ? lsb0 : msb0;
 
       if (debug_elaborate)
 	    cerr << get_fileline() << ": debug: Calculate element select "
 		 << name << "[" << *sel << "] element width " << elem_w
-		 << " from packed range [" << msb0 << ":" << lsb0 << "]." << endl;
+		 << " from " << (packed_layout ? "packed" : "unpacked")
+		 << " range [" << msb0 << ":" << lsb0 << "]." << endl;
 
 	// Constant index: fold to the selected element bits directly.
       if (const NetEConst*sel_c = dynamic_cast<NetEConst*>(sel)) {
@@ -6043,8 +6054,11 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 		  return res;
 	    }
 	    long sel_v = sel_c->value().as_long();
-	      // Canonical element index counting from the LSB end.
-	    long cidx = (msb0 >= lsb0) ? (sel_v - lsb0) : (lsb0 - sel_v);
+	      // Canonical element index counting from the LSB end of the flat
+	      // value.
+	    long cidx = packed_layout
+		  ? ((msb0 >= lsb0) ? (sel_v - lsb0) : (lsb0 - sel_v))
+		  : (sel_v - min_idx);
 	    long lsv = cidx * (long)elem_w;
 	    verinum result = param_part_select_bits(par_ex->value(), elem_w, lsv);
 	    NetEConst*res = new NetEConst(result);
@@ -6054,10 +6068,10 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 
 	// Variable index: build an indexed part select over the flat value.
 	// Normalize the outer index to a canonical element index, then scale
-	// it up to a bit offset by the element width (mirrors the outermost
-	// dimension handling in collapse_array_exprs()).
-      NetExpr*base = normalize_variable_base(sel, msb0, lsb0, elem_w,
-					     msb0 > lsb0);
+	// it up to a bit offset by the element width.
+      NetExpr*base = packed_layout
+	    ? normalize_variable_base(sel, msb0, lsb0, elem_w, msb0 > lsb0)
+	    : normalize_variable_base(sel, max_idx, min_idx, elem_w, true);
       base = scale_expr_by_const(base, elem_w);
 
       NetEConstParam*ptmp = new NetEConstParam(found_in, name, par_ex->value());
@@ -6088,18 +6102,35 @@ NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
 
       perm_string name = peek_tail_name(path_);
 
-	// A multi-dimension packed array parameter, e.g.
-	//   localparam logic [0:63][3:0] MEM = '{ ... };
-	// An index select MEM[idx] selects a whole element (a slice of the
-	// flat packed value that is the width of a single element), not a
-	// single bit. Handle that here before falling back to the single
-	// bit-select machinery used for ordinary packed vectors.
+	// An index select of an array parameter selects a whole element, not
+	// a single bit. This applies both to a multi-dimension packed array,
+	// e.g. localparam logic [0:63][3:0] MEM = '{...}, and to an unpacked
+	// array, e.g. localparam logic [3:0] MEM [0:63] = '{...} (whose value
+	// has been flattened to the packed-equivalent constant). The selected
+	// element keeps the remaining dimensions as its width.
       if (const netvector_t*par_vec = dynamic_cast<const netvector_t*>(par_type)) {
 	    if (par_vec->packed_dims().size() > 1) {
-		  NetExpr*res = elaborate_expr_param_slice_(par_ex, found_in,
-							    par_vec, sel, name);
-		  if (res) return res;
-		  return 0;
+		  const netranges_t&pdims = par_vec->packed_dims();
+		  long msb0 = pdims[0].get_msb();
+		  long lsb0 = pdims[0].get_lsb();
+		  unsigned long outer_w =
+			(msb0 >= lsb0) ? (unsigned long)(msb0 - lsb0 + 1)
+				       : (unsigned long)(lsb0 - msb0 + 1);
+		  unsigned long elem_w = par_vec->packed_width() / outer_w;
+		  return elaborate_expr_param_slice_(par_ex, found_in,
+						     msb0, lsb0, elem_w,
+						     true, sel, name);
+	    }
+      }
+      if (const netsarray_t*par_arr = dynamic_cast<const netsarray_t*>(par_type)) {
+	    if (!dynamic_cast<const netparray_t*>(par_arr)) {
+		  const netranges_t&udims = par_arr->static_dimensions();
+		  long msb0 = udims[0].get_msb();
+		  long lsb0 = udims[0].get_lsb();
+		  unsigned long elem_w = par_arr->element_type()->packed_width();
+		  return elaborate_expr_param_slice_(par_ex, found_in,
+						     msb0, lsb0, elem_w,
+						     false, sel, name);
 	    }
       }
 
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -5975,6 +5975,99 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
       return result;
 }
 
+/*
+ * Multiply a base expression by an unsigned constant, widening losslessly so
+ * the product cannot overflow. This mirrors the (file-static) make_mult_expr()
+ * in netmisc.cc, which is not exported.
+ */
+static NetExpr* scale_expr_by_const(NetExpr*expr, unsigned long val)
+{
+      unsigned val_wid = 0;
+      for (unsigned long tmp = val ; tmp > 1 ; tmp >>= 1)
+	    val_wid += 1;
+      unsigned use_wid = expr->expr_width() + val_wid + 1;
+
+      verinum val_v (val, use_wid);
+      val_v.has_sign(expr->has_sign());
+      NetEConst*val_c = new NetEConst(val_v);
+      val_c->set_line(*expr);
+
+      expr = pad_to_width(expr, use_wid, *expr);
+
+      NetEBMult*res = new NetEBMult('*', expr, val_c, use_wid, expr->has_sign());
+      res->set_line(*expr);
+      return res;
+}
+
+/*
+ * Handle an index select of a multi-dimension packed array parameter, e.g.
+ *
+ *    localparam logic [0:63][3:0] MEM = '{ ... };
+ *    assign o = MEM[idx];
+ *
+ * Unlike an ordinary packed vector, MEM[idx] selects a whole element (here a
+ * 4-bit slice), not a single bit. The parameter value is a flat packed
+ * constant, so this becomes a fixed-width part select of the flat value: for a
+ * constant index we fold it directly, otherwise we build an indexed part
+ * select over a NetEConstParam. Only the outermost dimension is indexed here
+ * (the common ROM/LUT case, `name[idx]`); the selected element keeps the
+ * remaining packed dimensions as its width.
+ */
+NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
+					      const NetScope*found_in,
+					      const netvector_t*par_vec,
+					      NetExpr*sel,
+					      perm_string name) const
+{
+      const netranges_t&pdims = par_vec->packed_dims();
+      long msb0 = pdims[0].get_msb();
+      long lsb0 = pdims[0].get_lsb();
+      unsigned long outer_w = (msb0 >= lsb0) ? (unsigned long)(msb0 - lsb0 + 1)
+					     : (unsigned long)(lsb0 - msb0 + 1);
+      unsigned long full_w = par_vec->packed_width();
+
+	// The element width is the flat width divided by the number of
+	// elements in the outermost dimension.
+      unsigned long elem_w = full_w / outer_w;
+
+      if (debug_elaborate)
+	    cerr << get_fileline() << ": debug: Calculate element select "
+		 << name << "[" << *sel << "] element width " << elem_w
+		 << " from packed range [" << msb0 << ":" << lsb0 << "]." << endl;
+
+	// Constant index: fold to the selected element bits directly.
+      if (const NetEConst*sel_c = dynamic_cast<NetEConst*>(sel)) {
+	    if (! sel_c->value().is_defined()) {
+		  NetEConst*res = new NetEConst(verinum(verinum::Vx, elem_w, true));
+		  res->set_line(*this);
+		  return res;
+	    }
+	    long sel_v = sel_c->value().as_long();
+	      // Canonical element index counting from the LSB end.
+	    long cidx = (msb0 >= lsb0) ? (sel_v - lsb0) : (lsb0 - sel_v);
+	    long lsv = cidx * (long)elem_w;
+	    verinum result = param_part_select_bits(par_ex->value(), elem_w, lsv);
+	    NetEConst*res = new NetEConst(result);
+	    res->set_line(*this);
+	    return res;
+      }
+
+	// Variable index: build an indexed part select over the flat value.
+	// Normalize the outer index to a canonical element index, then scale
+	// it up to a bit offset by the element width (mirrors the outermost
+	// dimension handling in collapse_array_exprs()).
+      NetExpr*base = normalize_variable_base(sel, msb0, lsb0, elem_w,
+					     msb0 > lsb0);
+      base = scale_expr_by_const(base, elem_w);
+
+      NetEConstParam*ptmp = new NetEConstParam(found_in, name, par_ex->value());
+      ptmp->set_line(found_in->get_parameter_line_info(name));
+
+      NetExpr*tmp = new NetESelect(ptmp, base, elem_w, IVL_SEL_IDX_UP);
+      tmp->set_line(*this);
+      return tmp;
+}
+
 NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
 					    const NetExpr*par,
 					    const NetScope*found_in,
@@ -5983,10 +6076,6 @@ NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
 {
       const NetEConst*par_ex = dynamic_cast<const NetEConst*> (par);
       ivl_assert(*this, par_ex);
-
-      long par_msv, par_lsv;
-      if(! calculate_param_range(*this, par_type, par_msv, par_lsv,
-				 par_ex->value().len())) return 0;
 
       const name_component_t&name_tail = path_.back();
       ivl_assert(*this, !name_tail.index.empty());
@@ -5998,6 +6087,25 @@ NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
       if (sel == 0) return 0;
 
       perm_string name = peek_tail_name(path_);
+
+	// A multi-dimension packed array parameter, e.g.
+	//   localparam logic [0:63][3:0] MEM = '{ ... };
+	// An index select MEM[idx] selects a whole element (a slice of the
+	// flat packed value that is the width of a single element), not a
+	// single bit. Handle that here before falling back to the single
+	// bit-select machinery used for ordinary packed vectors.
+      if (const netvector_t*par_vec = dynamic_cast<const netvector_t*>(par_type)) {
+	    if (par_vec->packed_dims().size() > 1) {
+		  NetExpr*res = elaborate_expr_param_slice_(par_ex, found_in,
+							    par_vec, sel, name);
+		  if (res) return res;
+		  return 0;
+	    }
+      }
+
+      long par_msv, par_lsv;
+      if(! calculate_param_range(*this, par_type, par_msv, par_lsv,
+				 par_ex->value().len())) return 0;
 
       if (sel->expr_type() == IVL_VT_REAL) {
 	    cerr << get_fileline() << ": error: Index expression for "
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -3551,8 +3551,17 @@ unsigned PECallFunction::elaborate_arguments_(Design*des, NetScope*scope,
 	    PExpr *tmp = args[idx];
 
 	    if (tmp) {
-		  parms[pidx] = elaborate_rval_expr(des, scope,
-						    def->port(pidx)->net_type(),
+		    // For a formal with unpacked dimensions the context type is
+		    // the array type; net_type() would give the element type
+		    // and the actual would then be rejected as needing an
+		    // index.
+		  const NetNet*formal = def->port(pidx);
+		  ivl_type_t ctype = formal->net_type();
+		  if (formal->unpacked_dimensions() > 0) {
+			if (const netarray_t*at = formal->array_type())
+			      ctype = at;
+		  }
+		  parms[pidx] = elaborate_rval_expr(des, scope, ctype,
 						    tmp, need_const);
 		  if (parms[pidx] == 0) {
 			parm_errors += 1;
--- a/elab_expr.cc
+++ b/elab_expr.cc
@@ -6082,6 +6082,153 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
       return tmp;
 }
 
+/*
+ * Add two non-negative bit-offset expressions, widening so the sum cannot
+ * overflow. (make_add_expr() in netmisc.cc is file-static and not exported.)
+ */
+static NetExpr* add_offset_exprs(NetExpr*a, NetExpr*b)
+{
+      unsigned w = a->expr_width();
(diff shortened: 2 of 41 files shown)


caryr:
FYI I believe the three failing tests you talked about in the description are because of a weird build dependency issue that I have not taken the time to figured out, but it can be worked around by doing a make clean in the vvp build directory and then rebuilding.

It looks like all your checks are failing in the CI because br1015b (perl VVP test) is now supported. The perl/python VVP tests and perl VPI tests all need to be 100% clean to pass the CI and then be accepted for merging.

I've not reviewed the code changes yet, but I've read the description and it describes a good approach.

We likely have a few other things we could discuss based on your website ;-), but I need to focus on getting the pull request queue cleaned up first.

marcobrambilla99:
Pushed `7c1bd3cd`, which should clear the CI.

You were right that `br1015b` was the cause. The detail that made it a hard failure rather than a
skip: it is listed twice, as `CE,-g2009` in `regress-ivl1.list` and as `normal,-g2009` in
`regress-sv.list`. `RegressionList.pm` keeps the first definition it sees
(`next if exists($testtype{$tname})`) and `vvp_reg.pl` loads `regress-ivl1.list` before
`regress-sv.list`, so the stale `CE` entry won and the correct entry never got a look in. Dropping
the ivl1 entry is enough.

I did the same in `regress-fsv.list`, and moved `br1015a` out of the "not supported in Icarus" block
while I was there. `br1015a` and `br1015b` are the same source; the only difference is that
`br1015b` gets `-g2009`. Since `--force-sv` rewrites every test to `-g2023`, `br1015a` passes now
too, so both run instead of being skipped. `br1015a` keeps its expected compile error in
`regress-vlg.list`, where the default generation still rejects it with
`Task/function port with unpacked dimensions requires SystemVerilog`.

```
vvp_reg.pl               3012 total, 3007 passed, 0 failed   (was 3006 / 1 failed)
vvp_reg.py                528 ran,     0 failed
vpi_reg.pl --with-pli1      85 total,   85 passed, 0 failed
vvp_reg.pl --force-sv     3020 total, 3001 passed, 0 failed, 19 NI (was 21 NI)
```

On the three `br_gh710` failures, there is nothing for you to chase: they were my build rather than a
dependency problem. Without zlib, `vvp` prints `LXT support disabled since zlib not available` on
stderr, and `vvp_reg.py` counts any stderr output as an error. A full scrub of the tree did not help;
configuring so zlib is found did. I will correct the description accordingly.

elab_expr.cc
```diff
@@ -5975,6 +5975,99 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
...
 
+/*
+ * Multiply a base expression by an unsigned constant, widening losslessly so
+ * the product cannot overflow. This mirrors the (file-static) make_mult_expr()
+ * in netmisc.cc, which is not exported.
+ */
+static NetExpr* scale_expr_by_const(NetExpr*expr, unsigned long val)
+{
+      unsigned val_wid = 0;
+      for (unsigned long tmp = val ; tmp > 1 ; tmp >>= 1)
```
caryr, line 5986: I prefer for single statements to be on the same line or like for this enclosed in a `{}` pair.

elab_expr.cc
```diff
@@ -5975,6 +5975,99 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
...
+NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
+					      const NetScope*found_in,
+					      const netvector_t*par_vec,
+					      NetExpr*sel,
+					      perm_string name) const
+{
+      const netranges_t&pdims = par_vec->packed_dims();
+      long msb0 = pdims[0].get_msb();
+      long lsb0 = pdims[0].get_lsb();
+      unsigned long outer_w = (msb0 >= lsb0) ? (unsigned long)(msb0 - lsb0 + 1)
```
caryr, line 6025: For the ternary operator I like to see the `:` at the end of the first line and then the second expression aligned to the first.

elab_expr.cc
```diff
@@ -5975,6 +5975,99 @@ static verinum param_part_select_bits(const verinum&par_val, long wid,
...
+					     : (unsigned long)(lsb0 - msb0 + 1);
+      unsigned long full_w = par_vec->packed_width();
+
+	// The element width is the flat width divided by the number of
+	// elements in the outermost dimension.
+      unsigned long elem_w = full_w / outer_w;
+
+      if (debug_elaborate)
```
caryr, line 6033: Missing the `{}` pair

caryr (changes requested):
Also the indenting looks off compared to the existing code. It is normally 6 spaces or a combination of tabs (expected to represent 8 spaces) ans spaces to give a multiple of 6. The comments are also normally indented two extra spaces relative to the code around them except for routine comments. (e.g. the first comment relative to code at 6 spaces is a tab (8 spaces).

elab_expr.cc
```diff
@@ -6011,29 +6019,32 @@ static NetExpr* scale_expr_by_const(NetExpr*expr, unsigned long val)
...
+ *     the element at index sel_v sits at bit (sel_v - lsb)*elem_w for an
+ *     ascending-LSB range, mirroring an ordinary packed part select.
+ *   - unpacked (packed_layout false): the value was flattened lowest-array-
+ *     index first, so the element at index sel_v sits at bit
+ *     (sel_v - min_index)*elem_w regardless of the declared range direction.
  */
 NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 					      const NetScope*found_in,
-					      const netvector_t*par_vec,
+					      long msb0, long lsb0,
```
caryr, line 6034: Should the range be passed in and then the MSB/LSB calculated inside the routine?

elab_expr.cc
```diff
@@ -6043,8 +6054,11 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 		  return res;
 	    }
 	    long sel_v = sel_c->value().as_long();
-	      // Canonical element index counting from the LSB end.
-	    long cidx = (msb0 >= lsb0) ? (sel_v - lsb0) : (lsb0 - sel_v);
+	      // Canonical element index counting from the LSB end of the flat
+	      // value.
+	    long cidx = packed_layout
```
caryr, line 6059: For multiple ternary I like to see the following:
```
(cond1) ? (cond2) ? true_for_cond2 :
                    false_for_cond2 :
          false_for_cond_1;
```

elab_expr.cc
```diff
@@ -6054,10 +6068,10 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
 
 	// Variable index: build an indexed part select over the flat value.
 	// Normalize the outer index to a canonical element index, then scale
-	// it up to a bit offset by the element width (mirrors the outermost
-	// dimension handling in collapse_array_exprs()).
-      NetExpr*base = normalize_variable_base(sel, msb0, lsb0, elem_w,
-					     msb0 > lsb0);
+	// it up to a bit offset by the element width.
+      NetExpr*base = packed_layout
```
caryr, line 6072: I prefer the `?` and `:` to be on the first line so you know the line is being continued.

elab_expr.cc
```diff
@@ -6088,18 +6102,35 @@ NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
...
 	    if (par_vec->packed_dims().size() > 1) {
-		  NetExpr*res = elaborate_expr_param_slice_(par_ex, found_in,
-							    par_vec, sel, name);
-		  if (res) return res;
-		  return 0;
+		  const netranges_t&pdims = par_vec->packed_dims();
+		  long msb0 = pdims[0].get_msb();
+		  long lsb0 = pdims[0].get_lsb();
+		  unsigned long outer_w =
+			(msb0 >= lsb0) ? (unsigned long)(msb0 - lsb0 + 1)
```
caryr, line 6117: See earlier ternary comments.

net_design.cc
```diff
@@ -499,6 +499,52 @@ void Design::evaluate_parameters()
...
+ * Returns nullptr if any element is not a foldable constant (e.g. a nested
+ * unpacked array), in which case the caller falls back to the normal error.
+ */
+static NetExpr* flatten_const_array_pattern(const NetEArrayPattern*pat,
+					    const netarray_t*arr,
+					    const LineInfo&loc)
+{
+      unsigned long elem_w = arr->element_type()->packed_width();
+      size_t count = pat->item_size();
+      if (elem_w == 0 || count == 0)
```
caryr, line 520: Make a single line instead of adding a `{}` pair

net_design.cc
```diff
@@ -499,6 +499,52 @@ void Design::evaluate_parameters()
...
+	    return nullptr;
+
+      verinum flat (verinum::Vx, elem_w * count, true);
+      for (size_t idx = 0 ; idx < count ; idx += 1) {
+	    const NetEConst*item = dynamic_cast<const NetEConst*>(pat->item(idx));
+	    if (!item)
```
caryr, line 526: Make a single line.

net_design.cc
```diff
@@ -499,6 +499,52 @@ void Design::evaluate_parameters()
...
+		  return nullptr;
+
+	    const verinum&iv = item->value();
+	    unsigned long base = (unsigned long)idx * elem_w;
+	    for (unsigned long bit = 0 ; bit < elem_w ; bit += 1) {
+		  verinum::V val;
+		  if (bit < iv.len())
```
caryr, line 533: Add `{}` pairs to the nested `if/else`

caryr (commented):
comments for the second commit

elab_expr.cc
```diff
@@ -6082,6 +6082,153 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
...
+	// stride[k] (in bits) = base_w * product of the inner-dimension sizes.
+      std::vector<unsigned long> stride(ndim);
+      unsigned long acc = base_w;
+      for (size_t k = ndim ; k-- > 0 ; ) {
+	    stride[k] = acc;
+	    acc *= udims[k].width();
+      }
+	// Width of the slice left after consuming nsel dimensions.
+      unsigned long result_w = base_w;
+      for (size_t k = nsel ; k < ndim ; k += 1)
```
caryr, line 6148: add a `{}` pair

elab_expr.cc
```diff
@@ -6082,6 +6082,153 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
...
+	    result_w *= udims[k].width();
+
+      if (debug_elaborate)
```
caryr, line 6151: Add a `{}` pair

elab_expr.cc
```diff
@@ -6082,6 +6082,153 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
...
+	    long hi = udims[k].get_msb();
+	    long mink = (lo <= hi) ? lo : hi;
+	    long maxk = (lo <= hi) ? hi : lo;
+
+	    if (NetEConst*idx_c = dynamic_cast<NetEConst*>(idx)) {
+		  if (!idx_c->value().is_defined()) {
+			out_of_bounds = true;
+		  } else {
+			long v = idx_c->value().as_long();
+			if (v < mink || v > maxk)
```
caryr, line 6178: Add a `{}` pair for the `if/else`

elab_expr.cc
```diff
@@ -6082,6 +6082,153 @@ NetExpr* PEIdent::elaborate_expr_param_slice_(const NetEConst*par_ex,
...
+							  stride[k], true);
+		  canon = scale_expr_by_const(canon, stride[k]);
+		  off_expr = off_expr ? add_offset_exprs(off_expr, canon) : canon;
+	    }
+      }
+
+	// All-constant index: fold to the selected element bits directly.
+      if (all_const) {
+	    if (out_of_bounds) {
+		  if (warn_ob_select)
```
caryr, line 6197: Add a `{}` pair

elab_expr.cc
```diff
@@ -6093,15 +6240,32 @@ NetExpr* PEIdent::elaborate_expr_param_bit_(Design*des, NetScope*scope,
...
+	//   localparam logic [3:0] TBL [0:3][0:15] = '{ '{...}, ... };
+	// Its value has been flattened row-major to a single packed constant.
+	// The path may apply one or more index expressions (a partial TBL[i] or
+	// the fully-indexed TBL[i][j]); each consumes one unpacked dimension and
+	// the selected slice keeps the remaining dimensions (plus the packed
+	// element) as its width. Handle this before the single-index machinery
+	// below, which only knows about the outermost dimension.
+      if (const netsarray_t*par_arr = dynamic_cast<const netsarray_t*>(par_type)) {
+	    if (!dynamic_cast<const netparray_t*>(par_arr)
+		&& par_arr->static_dimensions().size() > 1)
```
caryr, line 6256: Add a `{}` pair and I like to see the `&&` at the end of the first line to better indicate continuation of the line.

net_design.cc
```diff
@@ -499,34 +499,62 @@ void Design::evaluate_parameters()
...
+				   std::vector<const NetEConst*>&leaves)
+{
+      if (const NetEConst*c = dynamic_cast<const NetEConst*>(expr)) {
+	    leaves.push_back(c);
+	    return true;
+      }
+      if (const NetEArrayPattern*p = dynamic_cast<const NetEArrayPattern*>(expr)) {
+	    for (size_t idx = 0 ; idx < p->item_size() ; idx += 1) {
+		  const NetExpr*it = p->item(idx);
+		  if (!it || !collect_pattern_leaves(it, leaves))
```
caryr, line 519: Make a single line

net_design.cc
```diff
@@ -499,34 +499,62 @@ void Design::evaluate_parameters()
...
+ * caller falls back to the normal error.
  */
 static NetExpr* flatten_const_array_pattern(const NetEArrayPattern*pat,
 					    const netarray_t*arr,
 					    const LineInfo&loc)
 {
       unsigned long elem_w = arr->element_type()->packed_width();
-      size_t count = pat->item_size();
-      if (elem_w == 0 || count == 0)
+      if (elem_w == 0)
```
caryr, line 545: Make a single line

net_design.cc
```diff
@@ -499,34 +499,62 @@ void Design::evaluate_parameters()
...
+	    return nullptr;
+
+      std::vector<const NetEConst*> leaves;
+      if (!collect_pattern_leaves(pat, leaves))
```
caryr, line 549: Make a single line

net_design.cc
```diff
@@ -499,34 +499,62 @@ void Design::evaluate_parameters()
...
+	    return nullptr;
+      size_t count = leaves.size();
+      if (count == 0)
```
caryr, line 552: Make a single line

elab_expr.cc
```diff
@@ -3484,6 +3484,20 @@ NetExpr* PECallFunction::elaborate_base_(Design*des, NetScope*scope, NetScope*ds
 	   return value is the name within that scope. */
 
       if (NetNet*res = dscope->find_signal(dscope->basename())) {
+	      // A function that returns an unpacked array can be evaluated in a
+	      // constant context (its result is folded into a constant), but the
+	      // run-time code generator has no representation for returning a
+	      // whole unpacked array. Reject a run-time call with a clean
+	      // diagnostic rather than emitting something that cannot be drawn.
+	    bool res_is_uarray = res->unpacked_dimensions() > 0
```
caryr, line 3492: The `||` should be at the end of the first line.

elab_expr.cc
```diff
@@ -5292,9 +5306,23 @@ NetExpr* PEIdent::elaborate_expr(Design*des, NetScope*scope,
...
+
+	// When the whole net is referenced (no unpacked index), the type to
+	// compare against the context is the net's type as seen by the caller.
+	// For an unpacked array that is the array type, not the element
+	// (net_type()) type. This lets a whole unpacked array be used where an
+	// unpacked-array type is expected, e.g. `return r;` in a function that
+	// returns an unpacked-array typedef.
+      ivl_type_t net_ctype = net->net_type();
+      if (net->unpacked_dimensions() > 0 && use_comp.index.empty()) {
+	    if (const netarray_t*at = net->array_type())
```
caryr, line 5319: Add a `{}` pair or make it a single line.

elab_expr.cc
```diff
@@ -5292,9 +5306,23 @@ NetExpr* PEIdent::elaborate_expr(Design*des, NetScope*scope,
...
+		  net_ctype = at;
+      }
+
       ivl_type_t check_type = ntype;
       if (const netdarray_t*array_type = dynamic_cast<const netdarray_t*> (ntype)) {
-            if (array_type->type_compatible(net->net_type()) &&
+            if (array_type->type_compatible(net_ctype) &&
```
caryr, line 5325: The indenting of the following line looks wonky.

elab_expr.cc
```diff
@@ -5306,13 +5334,13 @@ NetExpr* PEIdent::elaborate_expr(Design*des, NetScope*scope,
...
 
-      if (! check_type->type_compatible(net->net_type())) {
+      if (! check_type->type_compatible(net_ctype)) {
 	    cerr << get_fileline() << ": error: the type of the variable '"
 		 << path_ << "' doesn't match the context type." << endl;
 
 	    cerr << get_fileline() << ":      : " << "variable type=";
-	    if (net->net_type())
-		  net->net_type()->debug_dump(cerr);
+	    if (net_ctype)
```
caryr, line 5342: Add a `{}` pair

emit.cc
```diff
@@ -458,13 +459,41 @@ void netclass_t::emit_scope(struct target_t*tgt) const
...
+ * A function that returns an unpacked array is evaluated entirely at
+ * elaboration time (in constant contexts); the run-time code generator has no
+ * representation for a whole-unpacked-array return value, and run-time calls to
+ * such a function are rejected during elaboration. So such a function scope is
+ * never referenced at run time and must not be emitted (drawing its body would
+ * fail). Detect that here.
+ */
+static bool skip_unpacked_array_func(const NetScope*scope)
+{
+      if (scope->type() != NetScope::FUNC)
```
caryr, line 472: Single line

emit.cc
```diff
@@ -458,13 +459,41 @@ void netclass_t::emit_scope(struct target_t*tgt) const
...
+	    return false;
+      const NetFuncDef*def = scope->func_def();
+      if (!def)
```
caryr, line 475: Single line

emit.cc
```diff
@@ -458,13 +459,41 @@ void netclass_t::emit_scope(struct target_t*tgt) const
...
+	    return false;
+      const NetNet*rsig = def->return_sig();
+      if (!rsig)
```
caryr, line 478: Single line

emit.cc
```diff
@@ -458,13 +459,41 @@ void netclass_t::emit_scope(struct target_t*tgt) const
...
+	    return false;
+	// The unpacked array may be recorded either as separate unpacked
+	// dimensions on the signal or as a netuarray_t net type, depending on
+	// how the return signal was created.
+      return rsig->unpacked_dimensions() > 0
```
caryr, line 483: `||` at the end of the first line.

emit.cc
```diff
@@ -458,13 +459,41 @@ void netclass_t::emit_scope(struct target_t*tgt) const
...
+}
+
 void NetScope::emit_scope(struct target_t*tgt) const
 {
       if (debug_emit) {
 	    cerr << "NetScope::emit_scope: "
 		 << "Emit scope " << scope_path(this) << endl;
       }
 
+      if (skip_unpacked_array_func(this))
```
caryr, line 494: Single line

tgt-vvp/draw_ufunc.c
```diff
@@ -40,8 +42,10 @@ static void function_argument_logic(ivl_signal_t port, ivl_expr_t expr)
 
 static void function_argument_real(ivl_signal_t port, ivl_expr_t expr)
 {
-	/* ports cannot be arrays. */
-      assert(ivl_signal_dimensions(port) == 0);
+	/* See function_argument_logic: array ports are handled in the send
+	   pass. */
+      if (ivl_signal_dimensions(port) > 0)
```
caryr, line 47: Single line

tgt-vvp/draw_ufunc.c
```diff
@@ -26,8 +26,10 @@ static void function_argument_logic(ivl_signal_t port, ivl_expr_t expr)
 {
       unsigned ewidth, pwidth;
 
-	/* ports cannot be arrays. */
-      assert(ivl_signal_dimensions(port) == 0);
+	/* An unpacked array port is copied straight from the source signal in
+	   the send pass, so there is nothing to evaluate onto the stack. */
+      if (ivl_signal_dimensions(port) > 0)
```
caryr, line 31: Single line

ca.log
```diff
@@ -0,0 +1,2 @@
+ca.sv: No such file or directory
```
caryr, line 1: What is this file? It should not be added

elab_expr.cc
```diff
@@ -3551,8 +3551,17 @@ unsigned PECallFunction::elaborate_arguments_(Design*des, NetScope*scope,
...
-		  parms[pidx] = elaborate_rval_expr(des, scope,
-						    def->port(pidx)->net_type(),
+		    // For a formal with unpacked dimensions the context type is
+		    // the array type; net_type() would give the element type
+		    // and the actual would then be rejected as needing an
+		    // index.
+		  const NetNet*formal = def->port(pidx);
+		  ivl_type_t ctype = formal->net_type();
+		  if (formal->unpacked_dimensions() > 0) {
+			if (const netarray_t*at = formal->array_type())
```
caryr, line 3561: Single line

elaborate.cc
```diff
@@ -4771,7 +4771,14 @@ NetProc* PCallTask::elaborate_build_call_(Design*des, NetScope*scope,
 	    NetExpr*rv = 0;
 
 	    if (args[parms_idx]) {
-		  rv = elaborate_rval_expr(des, scope, port->net_type(),
+		    // As for functions: an unpacked-array formal takes the
+		    // array type as its context, not the element type.
+		  ivl_type_t ctype = port->net_type();
+		  if (port->unpacked_dimensions() > 0) {
+			if (const netarray_t*at = port->array_type())
```
caryr, line 4778: Single line or `{}` pair

net_nex_input.cc
```diff
@@ -280,8 +280,10 @@ static void func_always_sens(NetFuncDef *func, NexusSet *result,
 	std::unique_ptr<NexusSet> in(new NexusSet);
 	for (unsigned idx = 0; idx < func->port_count(); idx++) {
 	      NetNet *net = func->port(idx);
-	      assert(net->pin_count() == 1);
-	      in->add(net->pin(0).nexus(), 0, net->vector_width());
+		// A port with unpacked dimensions has one pin per word, so
+		// remove them all, not just the first.
+	      for (unsigned pin = 0 ; pin < net->pin_count() ; pin += 1)
```
caryr, line 285: Add `{}` pair

caryr (commented):
fifth commit review

netmisc.cc
```diff
@@ -969,8 +969,17 @@ NetExpr* elab_and_eval(Design*des, NetScope*scope, PExpr*pe,
...
       if (dynamic_cast<const netuarray_t*>(lv_net_type)) {
-	    if (tmp->net_type())
-		  compatible = lv_net_type->type_compatible(tmp->net_type());
+	      // A whole unpacked-array reference (e.g. `return arr;` or an
+	      // array-to-array copy) elaborates to a NetESignal whose net_type()
+	      // is the element type. Compare against the signal's array type so
+	      // the unpacked array is accepted where an unpacked array is wanted.
+	    ivl_type_t rtype = tmp->net_type();
+	    if (const NetESignal*esig = dynamic_cast<const NetESignal*>(tmp)) {
+		  if (esig->word_index() == 0 && esig->sig()->array_type())
```
caryr, line 978: Add `{}` pair

eval_tree.cc
```diff
@@ -214,8 +221,15 @@ NetExpr* NetEBAdd::eval_arguments_(const NetExpr*l, const NetExpr*r) const
...
             unsigned wid = expr_width();
             ivl_assert(*this, wid > 0);
-            ivl_assert(*this, lval.len() == wid);
-            ivl_assert(*this, rval.len() == wid);
+	      // Constant folding during compile-time function evaluation can
+	      // deliver operands whose width does not match the node width
+	      // (e.g. an int loop variable used in a narrower index
+	      // expression). Resize to the node width; for the elaboration-time
+	      // callers the operands already match, so this is a no-op.
+	    if (lval.len() != wid)
```
caryr, line 229: Single line

eval_tree.cc
```diff
@@ -214,8 +221,15 @@ NetExpr* NetEBAdd::eval_arguments_(const NetExpr*l, const NetExpr*r) const
...
+		  lval = verinum(lval, wid);
+	    if (rval.len() != wid)
```
caryr, line 231: Single line

eval_tree.cc
```diff
@@ -171,8 +171,15 @@ NetExpr* NetEBAdd::eval_tree()
...
             unsigned wid = expr_width();
             ivl_assert(*this, wid > 0);
-            ivl_assert(*this, lval.len() == wid);
-            ivl_assert(*this, rval.len() == wid);
+	      // Constant folding during compile-time function evaluation can
+	      // deliver operands whose width does not match the node width
+	      // (e.g. an int loop variable used in a narrower index
+	      // expression). Resize to the node width; for the elaboration-time
+	      // callers the operands already match, so this is a no-op.
+	    if (lval.len() != wid)
```
caryr, line 179: Single line

eval_tree.cc
```diff
@@ -171,8 +171,15 @@ NetExpr* NetEBAdd::eval_tree()
...
+		  lval = verinum(lval, wid);
+	    if (rval.len() != wid)
```
caryr, line 181: Single line

emit.cc
```diff
@@ -512,6 +541,11 @@ bool NetScope::emit_defs(struct target_t*tgt) const
 		 << "Emit definitions for " << scope_path(this) << endl;
       }
 
+	// Skip functions that return an unpacked array (see
+	// skip_unpacked_array_func): they are only used at elaboration time.
+      if (skip_unpacked_array_func(this))
```
caryr, line 546: Single line

caryr (changes requested):
fifth commit review

caryr:
A bunch of mostly cosmetic changes requested. Not sure whay `ca.log` was added, that looks wrong and it looks like you can pass in the range instead of the lsb/msb, but I may have missed something since this is a large set of changes.

PR #1472, opened 2026-08-16
Use `symbol_search()` for task call and `disable` target lookup

larsclausen:
Task and function calls used as statements currently use `Design::find_task()` and `Design::find_function()`, which only search for scopes of the requested type. A closer variable, named event, or parameter is skipped. This can call a hidden task or function, and can make the non-void discarded-return path resolve a different symbol and abort.

The LRM section 26.3 requires function and task call lookup to consider all identifiers visible through the end of the current scope. Perform one `symbol_search()` and use its result for both the task lookup and the function fallback. Accept a function scope or the return variable of a recursive function call, while preserving the existing method fallback. Remove the now unused `Design::find_task()` and `Design::find_function()` helpers.

Similarly `disable` target search uses `Design::find_scope()`, which searches only scope names. A task or named block from an outer scope is therefore selected even when a variable, named event, or parameter in a closer scope has the same name. Also convert that to `symbol_search()`


larsclausen:
Tests will pass once rebased onto #1473

PR #1478, opened 2026-08-17
Support hierarchy identifiers shadowing type identifiers

larsclausen:
The lexer returns `TYPE_IDENTIFIER` when an identifier matches a visible typedef. It does not know about hierarchy scopes declared later, so an imported type can prevent a forward-referenced generate block with the same name from being used as the first hierarchy component:
```SystemVerilog
    package p;
      typedef int SCOPE;
    endpackage
    import p::*;
    module test;
      initial SCOPE.value = 1;
      if (1) begin : SCOPE
        int value;
      end
    endmodule
```
The same applies to inherited class properties, which are not visible to the parser:
```SystemVerilog
    typedef int value;
    class Base;
      int value;
    endclass
    class Derived extends Base;
      function int get;
        return value;
      endfunction
    endclass
```
Accept `TYPE_IDENTIFIER` as the first hierarchy component and build it through the same path as `IDENTIFIER`. This applies the existing hierarchy-index validation before lookup and keeps malformed components out of elaboration.

Named types in expression-like contexts are already resolved through identifier elaboration, so remove their separate `type_value` alternatives and use the common hierarchy expression path. Keep atomic types in `type_value`. This removes the grammar overlap without adding dedicated parameter and system argument rules.

These identifiers can now reach continuous assignment l-value elaboration. Use the common type diagnostic there. The implicit-net lookup checks typedefs, so invalid type uses in continuous assignments, gate terminals, and module port connections remain rejected rather than silently becoming nets.

This only permits hierarchy names to shadow visible type names. A hierarchical path that resolves to a type remains rejected during elaboration.


caryr:
Great work on cleaning up all the corner cases and getting the base parser to handle things properly! This work is a good example of small AI assisted steps to improve the tool.

My intuition guessed that the `type` operator would allow a hierarchical reference to get the type of a non-local expression so that a generate case using the type could adapt to changes in other hierarchies, but that is actually forbidden. I do read the descriptions of your pull requests, though I have not been reviewing the code lately. Too many other reviews to work on when I have some free time. Which has been scarce this summer.

One thing you could help with is should we have a standard AI file that specifies formatting? I've noticed the code being submitted often doesn't handle things the way I like to see them. Basically a style guide for the AIs. For example spacing expectations for code/comments, line continuation, block pair `{}` usage, ternary formatting, etc. An AI should be able to extract this from the code and adapt as needed, but I can review the rules it thinks are there or to describe exceptions versus what it could see as multiple styles because non-conformant code has been accepted previously.

We should likely create a discussion for this so everything is available. Let me know if you have time to help with this.

larsclausen:
> Great work on cleaning up all the corner cases and getting the base parser to handle things properly! This work is a good example of small AI assisted steps to improve the tool.
> 

I've started this a long time ago and had a whole bunch of unfinished branches and there always just more corner cases. But have been making good progress now and once this PR is merged there is not too much left to do. The goal is to remove the requirement that the parser needs to know is and what isn't a type identifier altogether. There are some cases with class parameters where it is not known until elaboration time if a identifier inside class is a type or something else.

> One thing you could help with is should we have a standard AI file that specifies formatting? I've noticed the code being submitted often doesn't handle things the way I like to see them. Basically a style guide for the AIs. For example spacing expectations for code/comments, line continuation, block pair `{}` usage, ternary formatting, etc. An AI should be able to extract this from the code and adapt as needed, but I can review the rules it thinks are there or to describe exceptions versus what it could see as multiple styles because non-conformant code has been accepted previously.
> 
> We should likely create a discussion for this so everything is available. Let me know if you have time to help with this.

I can look into creating an AGENTS.md with style guidelines. Agents are generally good a picking up the style from a project, but we do not always very consistent with the existing formatting. There is also a lot of old-style C++ which shouldn't be used by new code. E.g. 0 vs nullptr.

One thing I tried a while ago is to put together a `.clang-format` so we have consistent formatting throughout. But the current formatting style is pretty unique and I think the closest I got was still re-formatting around 25% of the existing lines. We could do that one time if we really wanted to and then enforce clang format for any new code.

PR #1492, opened 2026-09-07
Disambiguate named declarations and module instances

larsclausen:
Commit 2d498363d7f3("Handle instance array of port-less modules.") added support for arrayed instances of modules with no ports without port-connection parentheses, e.g. `T x[1:0];`. This is an Icarus extension, not valid Verilog or SystemVerilog: IEEE 1800-2023 23.3.2 requires `T x[1:0]();` even for a module with no ports.

This conflicts with the SystemVerilog array variable declaration `T x[1:0];`. The parser currently relies on the lexer returning `TYPE_IDENTIFIER` for a visible type name, rather than `IDENTIFIER`, to distinguish the two interpretations.

In preparation for removing `TYPE_IDENTIFIER`, keep the shared prefix in one grammar rule and select the interpretation in its actions:

1) Parenthesized connections or parameter overrides select an instance, e.g. `T x();` or `T #(1) x();`. Connections in a later entry also select an instance list, as in `T a[1:0], b();`. Reject scalar entries without parentheses, e.g. `T a[1:0], b, c();`. Continue to reject invalid variable initializers on instances, e.g. `T a[1:0] = 0, b();`.

2) When packed dimensions precede the second identifier, interpret the item as a variable declaration, e.g. `T [1:0] x;`.

3) Without connections, interpret a list containing a scalar entry or an initializer as variable declarations, e.g. `T x;` or `T x[1:0] = 0;`. This does not add support for scalar module instances without parentheses.

4) For the remaining array-only forms without initializers, such as `T x[1:0];`, look up the leading name. A visible type selects a variable declaration; otherwise build no-port module instances.

The remaining lookup in case 4 is not ideal: it requires the parser to keep tracking type names to preserve the parentheses-less instance extension. However, this conflict occurs only in module-item contexts, including generate blocks, with an unqualified leading name. In these scopes, visible declarations and imports let us determine at parse time whether the name denotes a type, without resolving its full definition.

For variable declarations, build an unresolved named type and leave type validation to elaboration. Unlike the ambiguous module-item case, an inherited type name inside a class may depend on a type-parameter base class and only become resolvable at elaboration. The same applies to future class-qualified declarations such as `C::T x[1:0];`, which must be interpreted as declarations without a parse-time lookup of `C::T`.


caryr:
I'm not against restricting the syntax or requiring a flag to enable the alternate syntax. Much of the extensions are because they are useful or earlier versions of the standard did not completely define the syntax. I'm not certain this falls into that category. We should not enshrine extensions just because they currently exist when it's entirely possible it was not a deliberate choice and just an artifact of a design that was not fully thought out. I know it's more work to dig into the code to look for clues regarding why this was implemented as it is, but it may be useful to help get more standard compliant parsing.

Just an editorial comment proceed as you see fit!

larsclausen:
I though about adding a flag, but even if we introduce a flag, we still need to keep the parser compatible. We do have a regression test that explicitly tests for this atm, so that's why I kept it for now. Based on the commit that introduced it this was deliberate, not just a mistake https://github.com/steveicarus/iverilog/commit/2d498363d7f3f62d1bdf46dce76fe2a5f073551f

We should probably still add a flag, together with some other things that are iverilog extensions. But let me finish the type_identifier removal first. Very close now.

PR #1500, opened 2026-09-10
Implement toupper() and tolower() string functions

msinger:
Hi, I have implemented the string case conversion functions `toupper()` and `tolower()`. I'm not familiar with the code base and I'm not sure if this is okay or if something is missing. Also I don't know if the `regress-*.list` file I used is the correct one for this type of test. Please let me know if I messed up.


ivtest/vvp_tests/sv_string_toupper_tolower.json
```diff
@@ -0,0 +1,5 @@
+{
+    "type"          : "normal",
+    "source"        : "sv_string_toupper_tolower.v",
+    "iverilog-args" : [ "-g2005-sv" ]
+}
```
larsclausen, line 9: This need an exception for vlog95 backend that it will fail because strings are not supported.  Similar to e.g. https://github.com/steveicarus/iverilog/blob/master/ivtest/vvp_tests/array_slice_concat.json#L4-L7

msinger: Thank you!

I wasn't aware that I can run the vvp tests for vlog95. I think the README.txt needs some updates.

caryr:
FYI I pushed support for toupper/tolower on parameter values and checks that all the string methods are called with the proper number of arguments.

